PR #1, opened 2016-01-13
Allow more than one option to be a colour picker

jesjimher:
Instead of checking on full name (id=="colour"), it now checks on a prefix (id starts with "colour"). Needed for visualizations that may define more than one colour (for example stacked).

Without this change, PR #444 (colour configuration in Stacked visualization) doesn't work inside dashboards.


glynhudson:
Thanks! Works great. Very sorry it has taken so long to merge this

PR #2, opened 2016-04-02
Estonian language translation for dashboard

kkaasan:
Could someone point out, where are main menu language files located?


glynhudson:
@chaveiro can you help? Have you worked on the language translation? Cc @TrystanLea

glynhudson:
@kkaasan I have just tried to test your translations. However when I select Estonian (or any other language in my account setting nothing seems to chance, all text is still in English. I have tried logging out then back in again. Is there another step required to enable alternative foreign translations? Could it be my browser is set to English?

chaveiro:
The path is defined at the *_menu.hp files. 
    $domain = "messages";
    bindtextdomain($domain, "Modules/dashboard/locale");
    bind_textdomain_codeset($domain, 'UTF-8');
See https://github.com/emoncms/dashboard/blob/master/dashboard_menu.php#L20

kkaasan:
Hi!

Had the same problem and then found this manual: [https://github.com/emoncms/emoncms/blob/master/docs/gettext.md#4-install-gettext](https://github.com/emoncms/emoncms/blob/master/docs/gettext.md#4-install-gettext)

Made:

`sudo locale-gen et_EE.UTF-8`
`sudo /etc/init.d/apache2 restart`

and after that the language changed.

glynhudson:
Thanks, however I can't seem to get the locals to generate. 

I have changed that there are multiple entries for `et` in supported and `gettext` is installed  

```
cat  /usr/share/i18n/SUPPORTED | grep et
et_EE.UTF-8 UTF-8
et_EE ISO-8859-1
et_EE.ISO-8859-15 ISO-8859-15
```

When try and generate the locals for `et` it seems to only generate `gb` instead:

```
sudo locale-gen et_EE.UTF-8
Generating locales (this might take a while)...
  en_GB.UTF-8... done
Generation complete.
```

I'm testing on a Raspberry Pi running Raspbian Jessie. Can you see what I'm missing?

PR #9, opened 2016-09-05
Add key bindings to move things around the dashboard editor

mattjgalloway:
This adds key bindings when on the dashboard editor so that you can move items around with the arrow keys. Items move by 1 grid size unit in whichever direction of arrow you press.

### DIFF
--- a/Views/js/designer.js
+++ b/Views/js/designer.js
@@ -511,6 +511,8 @@ var designer = {
 
                 designer.draw();
                 designer.modified();
+
+                e.preventDefault();
             }
         });
 
--- a/Views/js/designer.js
+++ b/Views/js/designer.js
@@ -483,6 +483,9 @@ var designer = {
             if (keyCode >= 37 && keyCode <= 40) {
                 if (!designer.selected_box) return;
 
+                var targetTagName = e.target.tagName.toLowerCase();
+                if (targetTagName === 'input' || targetTagName === 'textarea') return;
+
                 var left_shift = 0;
                 var top_shift = 0;
 
--- a/Views/js/designer.js
+++ b/Views/js/designer.js
@@ -476,6 +476,44 @@ var designer = {
             }
         });
 
+        // Key events
+        window.addEventListener('keydown', function(e) {
+            var keyCode = e.keyCode;
+
+            if (keyCode >= 37 && keyCode <= 40) {
+                if (!designer.selected_box) return;
+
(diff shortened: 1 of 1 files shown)


Paul-Reed:
I was about to comment on that!!
Thanks

Paul-Reed:
It's still not happening for me, I have your latest version of designer.js and the arrow keys have no effect other than move the page up or down a little.
Using windows 10, tried Firefox, Chrome & Explorer (current versions), clearing their caches, even trying incognito windows to no avail.

mattjgalloway:
@Paul-Reed - Hmm that's weird! I'm mainly using Chrome, but it also works for me in Safari. How odd! I'm using a Mac, but I can't believe that would make a difference here. Unless the key code is different for the arrow keys...

What does this website say when you press the arrow keys on your keyboard?
http://keycode.info/

Paul-Reed:
37, 38, 39 & 40, identical to your commit.

mattjgalloway:
@Paul-Reed - can you try with that latest commit please? I switched it to binding through jquery which may standardise things across browsers/OSes for us.

Paul-Reed:
The jquery switch on your last commit has now enabled the arrow keys. and appear to be now working as intended. Great!
I'll do a little more testing, but it's looking good, and especially as it nudges the grid distance which is now user selectable.

mattjgalloway:
Yeh, I found this incredibly useful. Moving things around by dragging the centre is rather frustrating especially when the centre isn't very visible (thus hard to find) due to the colour of the widget.

Paul-Reed:
Thanks Matt, works fine now provided the cache is refreshed to clear the javascript. No errors, and no detrimental issues for android/ios.
Two very much appreciated merges :+1:

Paul-Reed:
Matt, one thing that I have found;
If moving objects with a mouse, it was extremely difficult to move objects off of the page, so they could always be recovered.
However, using the arrow keys, objects can be moved off-page, and if focus is lost they cannot be retrieved/deleted.
Could restrictions be applied to stop this?

Paul

mattjgalloway:
Ah yes. That'll need fixing. Good spot. I'll work on that ASAP.

I've always wondered why the page isn't infinite anyway, and just scroll as far as is needed to view every element. But maybe there's a reason why that is hard to implement.

mattjgalloway:
@Paul-Reed - fixed with #10.

PR #10, opened 2016-09-06
Ensure that the element doesn't move off the page when using arrow keys

mattjgalloway:
This fixes the issue that @Paul-Reed noticed, whereby you can move elements off the screen with the arrow keys.

I've implemented it in the same way as the mouse events work. That is, you can't move the centre of the elements off the left, top or right of the page. And when you move the element off the bottom of the page, the page height simply changes to accommodate the element with an extra `grid_size` amount as padding.

### DIFF
--- a/Views/js/designer.js
+++ b/Views/js/designer.js
@@ -515,8 +515,27 @@ var designer = {
 
                 var snap_amount = Math.max(designer.grid_size, 1);
 
+                // Check that the element doesn't go off the page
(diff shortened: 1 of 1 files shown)


Paul-Reed:
Thanks Matt for the quick response. I'm out till this evening, but I'll be in touch later.

Paul-Reed:
Thanks Matt

PR #11, opened 2016-09-06
Add buttons to move elements forwards and backwards

mattjgalloway:
This adds buttons to move the currently selected element forwards and backwards.

[As requested by @Paul-Reed ;-). But a really neat idea, which is highly useful!]


Paul-Reed:
Works really well! and presents an opportunity for everyone now to be able to re-order their dashboard items so that underlying widgets etc can be accessed. I previously re-arranged mine by editing the MYSQL table, but this is way slicker!
Very nice clear code, thanks Matt.

glynhudson:
Awesome! Thanks.

- sent from my mobile device

PR #12, opened 2016-09-06
Add an option to feedvalue to put the unit at the front

mattjgalloway:
Useful for £/$ units.

screen shot 2016-09-06 at 16 18 39

### DIFF
--- a/widget/feedvalue/feedvalue_render.js
+++ b/widget/feedvalue/feedvalue_render.js
@@ -47,9 +47,15 @@ function feedvalue_widgetlist()
         [6,    "6"]
     ];
 
-  addOption(widgets["feedvalue"], "feedid",     "feedid",  _Tr("Feed"),     _Tr("Feed value"),      []);
-  addOption(widgets["feedvalue"], "units",      "value",   _Tr("Units"),    _Tr("Units to show"),   []);
-  addOption(widgets["feedvalue"], "decimals",   "dropbox", _Tr("Decimals"), _Tr("Decimals to show"),    decimalsDropBoxOptions);
+  var unitEndOptions = [
+        [0, "Back"],
+        [1, "Front"]
+    ];
+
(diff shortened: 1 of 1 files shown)


Paul-Reed:
Unit position dropdown is not visible in config.
feedvalue

mattjgalloway:
@Paul-Reed - can you try refreshing? It really should be there, and shows in mine as per screenshot. I've found sometimes you need to refresh a couple of times to make sure everything is updated and not from old cached objects.

Paul-Reed:
I've already ruled that out. Tried 3 different browser with cache cleared, tried incognito window, and also connected by local IP address - same result.

mattjgalloway:
Well that's fun then! Hmm. I'll try and see if I can figure out why - but the code is there to show that text and the box :-/.

mattjgalloway:
@Paul-Reed - I'm totally at a loss. I've checked out this specific branch and it still works for me.

Maybe you could try putting a breakpoint in `widget/feedvalue/feedvalue_render.js` to verify you are indeed hitting the refreshed code when you're testing this?

Paul-Reed:
OK, I've just re-installed dashboard from scratch, this was the sequence;

- Delete existing 'dashboard' directory
- Git clone your dashboard fork
- Rename the directory from 'emoncms-dashboard' to 'dashboard'
- Checkout your 'feedvalue_unit_position branch'
- Git pull - up to date
- Reboot raspberry pi
- Clear browser cache & reload emoncms

Still same result, even tried new mobile devices.

I'll have more time tomorrow night to explore further.

mattjgalloway:
Ugh that's really, really odd. I hope it's something that will magically fix itself for you tomorrow, because I've no idea why it wouldn't be showing it :-(. Thanks for trying!

Paul-Reed:
Sorted! I found another version of feedvalue_render.js on my pi, which I was being picked up instead of the dashboard/widget/feedvalue version.
All good now, no errors and a helpful addition.
Thanks Matt

PR #13, opened 2016-09-07
Add binding on the delete key to delete selected boxes

mattjgalloway:
As title. You can now press the backspace or delete key to delete items.

In this PR I also pull out the arrow key logic into a separate function. Gotta keep things sane in this file!

### DIFF
--- a/Views/js/designer.js
+++ b/Views/js/designer.js
@@ -355,6 +355,17 @@ var designer = {
         designer.modified();
         designer.edit_mode = true;
     },
+    
+    'delete_selected_boxes': function(){
+        if (designer.selected_box) {
+            delete designer.boxlist[designer.selected_box];
+            $("#"+designer.selected_box).remove();
+            designer.selected_box = 0;
+            designer.draw();
+            designer.modified();
+            $("#when-selected").hide();
+        }
+    },
 
     'get_unified_event': function(e){
         var coors;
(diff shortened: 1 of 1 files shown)


Views/js/designer.js
```diff
@@ -494,62 +570,27 @@ var designer = {
...
+                        e.preventDefault();
+                    }
+                    break;
+                case 16: // Shift
+                    designer.shiftdown = true;
+                    break;
+                case 8:
+                case 46: // Backspace & delete
+                    if (designer.handle_delete_key_event(e)) {
+                        e.preventDefault();
```
Paul-Reed: Matt, the e.preventDefault(); doesn't appear to function for case 8 (backspace).
In Windows, if the backspace key is hit, the default action of restoring the previous page occurs and the item is not deleted.
This isn't a big issue, as most Windows users would I assume use the delete key anyway.

Paul

Paul-Reed:
Matt, the e.preventDefault(); doesn't appear to function for case 8 (backspace).
In Windows, if the backspace key is hit, the default action of restoring the previous page occurs and the item is not deleted.
This isn't a big issue, as most Windows users would I assume use the delete key anyway.

Paul

mattjgalloway:
Interesting. I guess that's OK though, if delete works for Windows.

Paul-Reed:
Thanks Matt, looks a lot cleaner now.

PR #14, opened 2016-09-08
Fix dial widget start limit issues

mattjgalloway:
Me again!

I noticed 2 issues with the `dial` widget:

  1. screen shot 2016-09-08 at 12 10 15
  Umm, that doesn't look right. The start limit is wrong. It should say `-2000` here. Notice that the needle is in the right position, it's just the limit that's wrong.

  2. screen shot 2016-09-08 at 12 10 30
  Also not right. When the dial is of a type where it's not zero at the far left, the needle can overshoot on the left.

This PR fixes both of those issues, and adds some documentation in the code to help understand the `needle` calculation. I found myself having to dissect that while figuring out how to fix these issues, and so I thought I'd be kind for the next person who comes along.

Here's a couple of screenshots showing how the issue is now fixed.

screen shot 2016-09-08 at 12 57 18 screen shot 2016-09-08 at 12 57 26

### DIFF
--- a/widget/dial/dial_render.js
+++ b/widget/dial/dial_render.js
@@ -150,16 +150,12 @@ function draw_gauge(ctx,x,y,width,height,position,maxvalue,units,type, offset, g
   if (type == 0){ // Standard dial from 0 to maxvalue if offset is not set.
     // TODO: seperate between needle position at maximum/minimum and the value displayed.
     // TODO: Do we need to limit the value being displayed? Only needle position should be limited.
-    if (position<0)
-      position = 0;
   }
   else if (type == 1){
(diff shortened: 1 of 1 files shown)


Paul-Reed:
Thanks yet again Matt.
I've been aware of this issue for some time, so it's good that it's now fixed.

Paul

mattjgalloway:
@Paul-Reed - great stuff! If you see other issues (or know of any now), may I suggest opening an issue for it? It would be great to see issues being created so that people can pick things off as and when. I'm only finding things when I spot them, and I'm finding things as I delve in further to OEM, but I'm certain there's loads of areas I'm missing!

PR #15, opened 2016-09-08
Don't do the delete if you're inside an input or textarea

mattjgalloway:
@Paul-Reed - sorry about this... I just noticed that with my delete key binding, you can end up deleting the selected element while you're configuring it, if you press backspace / delete while in a text box. This commit fixes it.


Paul-Reed:
Well spotted!

glynhudson:
Thanks!

- sent from my mobile device

PR #16, opened 2016-09-09
Add undo/redo

mattjgalloway:
This adds undo/redo handling to the dashboard.

Looking for thought on this before it gets committed. The menu bar is getting a bit clogged with these and with the move forward/backward I added before. Maybe we should split into 2 rows? Have undo/redo and save on a second row?

### DIFF
--- a/Views/js/designer.js
+++ b/Views/js/designer.js
@@ -38,6 +38,10 @@ var designer = {
 
     'mousedown': false,
 
+    'undostack': [],
+    'redostack': [],
+    'nextundostate': null,
+
     'init': function(){
         designer.cnvs = document.getElementById("can");
         designer.ctx = designer.cnvs.getContext("2d");
@@ -47,6 +51,7 @@ var designer = {
         designer.draw();
         designer.widget_buttons();
         designer.add_events();
+        designer.check_undo_state();
     },
 
 
@@ -60,6 +65,78 @@ var designer = {
 
     'modified': function(){
         $("#save-dashboard").attr('class','btn btn-warning').text(_Tr("Changed, press to save"));
+    },
+
+    'start_save_undo_state': function(){
+        if (designer.nextundostate !== null) {
+            console.log("Imbalanced undo state save start/end!");
+        }
+        var newstate = $("#page").html();
+        designer.nextundostate = newstate;
+    },
+
+    'end_save_undo_state': function(){
+        if (designer.nextundostate === null) {
+            console.log("No undo state to save!");
+            return;
+        }
+        var currentstate = $("#page").html();
+        if (currentstate === designer.nextundostate) {
+            designer.cancel_save_undo_state();
+        } else {
+            designer.undostack.push(designer.nextundostate);
+            designer.redostack = [];
+            designer.nextundostate = null;
+            designer.check_undo_state();
+        }
+    },
+
+    'cancel_save_undo_state': function(){
+        designer.nextundostate = null;
+    },
+
+    'undo': function(){
+        if (designer.undostack.length == 0) return;
+
+        var currentstate = $("#page").html();
+        var laststate = designer.undostack.pop();
+
+        designer.redostack.push(currentstate);
+
+        $("#page").html(laststate);
+        designer.selected_box = 0;
+        designer.scan();
+        designer.draw();
+        designer.check_undo_state();
+    },
+
+    'redo': function(){
+        if (designer.redostack.length == 0) return;
+
+        var currentstate = $("#page").html();
+        var laststate = designer.redostack.pop();
+
+        designer.undostack.push(currentstate);
+
+        $("#page").html(laststate);
+        designer.selected_box = 0;
+        designer.scan();
+        designer.draw();
+        designer.check_undo_state();
+    },
+
+    'check_undo_state': function(){
+        if (designer.undostack.length > 0) {
+            $("#undo-button").prop('disabled', false);
+        } else {
+            $("#undo-button").prop('disabled', true);
+        }
+
+        if (designer.redostack.length > 0) {
+            $("#redo-button").prop('disabled', false);
+        } else {
+            $("#redo-button").prop('disabled', true);
+        }
     },
 
     'onbox': function(x,y){
@@ -84,10 +161,12 @@ var designer = {
     },
 
     'scan': function(){
+        var seenboxes = [];
         for (z in widgets){
             $("."+z).each(function(){
                 var id = 1*($(this).attr("id"));
                 if (id>designer.boxi) designer.boxi = id;
+                seenboxes.push(id);
                 designer.boxlist[id] = {
                     'top':parseInt($(this).css("top")),
                     'left':parseInt($(this).css("left")),
@@ -103,6 +182,14 @@ var designer = {
                 if ((designer.boxlist[id]['top'] + designer.boxlist[id]['height'])>designer.page_height) designer.page_height = (designer.boxlist[id]['top'] + designer.boxlist[id]['height']);
             });
         }
+
+        var allboxes = Object.keys(designer.boxlist);
+        $.each(allboxes, function(i, box){
+            box = 1 * box;
+            if ($.inArray(box, seenboxes) < 0) {
+                delete designer.boxlist[box];
+            }
+        });
     },
     
     // given an element and a style name, returns the exact style value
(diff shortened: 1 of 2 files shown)


Views/js/designer.js
```diff
@@ -103,6 +178,14 @@ var designer = {
...
             });
         }
+
+        var allboxes = Object.keys(designer.boxlist);
+        $.each(allboxes, function(i, box){
+            box = 1 * box;
+            if ($.inArray(box, seenboxes) < 0) {
+                delete designer.boxlist[box];
+            }
+        });
```
mattjgalloway: This needs to be done so that boxes that used to be there, which aren't now (e.g. add widget, undo), are removed.

Paul-Reed:
Hhmm, I see what you mean about the menu bar starting to clog up. Trouble is they are all useful and welcome features!
Instead of splitting into 2 lines, would there be value in cascading the functions, similar to html menu items. So for example, have one button called 'Layers' which when selected cascades 2 options 'Up' and 'Down'. Similarly, have a 'Edit' button which cascades 'Undu' and 'Redo'.
Would this be viable?
Although this would mean 2 clicks instead of 1, they would then follow the same style & intuitive access as the other dashboard functions, and keep the menu looking clean.

Paul

mattjgalloway:
> Instead of splitting into 2 lines, would there be value in cascading the functions, similar to html menu items. So for example, have one button called 'Layers' which when selected, cascades 2 options - 'Up 🔼' and 'Down 🔽'. Similarly, have a 'Edit' button which cascades - 'Undu ↪️' and 'Redo ↩️'.
> Would this be viable?

It would be possible, but my concern would be that undo/redo/delete/configure I think should be a minimum number of clicks away, and 1 as it is currently is ideal. They're things you do a lot, so minimising clicks to use them is important.

I don't think these options should go into a drop down. An alternative would be purely glyphs and no text on the buttons. That would make them shorter.

Paul-Reed:
The undo/redo functions could be glyphs as they are readily identifiable, not so sure about layers...
I see that the configure/forward/backward/delete menus only appear when a object is selected. At the same time could the text/containers/widgets/visualizations disappear,  so we have a fully responsive menu?
This would create some space...

Paul

TrystanLea:
Wow this is super nice! and the forward/backwards feature! shall I go ahead and merge?

TrystanLea:
Thanks for all the work your doing Matt! it looks excellent!

mattjgalloway:
I'd be happy for a merge,  but we do need to solve the too many buttons issue. I'm focussing on th iOS app at the moment but happy to take a look at something on this front soon.

WDYT about glyphs only @TrystanLea? That would clear up space.

TrystanLea:
How about icons only but with hover over tooltip? forward/back might be less clear without text helper..

Paul-Reed:
> How about icons only...

Is that to be replicated throughout the emoncms structure? there surely needs to be consistency.
Couldn't this be accommodated by using responsive menus?

Paul

mattjgalloway:
Agreed on consistency. Although undo/redo icons are pretty common these days.

I'm against hiding things under a menu, simply because these things need to be single click access IMO.

I think hiding the widget/visualisation buttons when an item is selected isn't great because you then have to deselect in order to add a new widget. Any friction like that can get tedious in this sort of thing I would suggest.

I could try the 2 lines option. That might work. To be fair we are already quite crowded on medium sized screens anyway.

Paul-Reed:
I don't that the 2 lines option is the way to go, as it eats into the workspace and is aesthetically poor.
Perhaps a compromise?
If 'Visualisation' is shortened to 'Vis' & 'Forward - Backward' becomes 'Up - Down' (which I think it should be anyway), this would create a little more space to accommodate 2 glyphs - 'Undo - Redo' with tooltips.

Paul

firefox7518:
Would it be an option to actually have a floating box instead of the fix position. I'm always a little bit different to others and as I'm working on 4K Monitors I always hated the fixed position to edit dashboards for 4K dashboards :-O 

image
I left is as a line but we could easily do a box like in photoshop or gimp!? What do you think?

Paul-Reed:
Nice suggestion Andreas, it works well for Photoshop.
I think we could learn a lot from parallel systems.

Paul

Paul-Reed:
I think that the only way that we can evaluate contributions is through pull-requests, we can't really dive into other peoples forks and pull out new features. That's not the way that the system works.

Paul

firefox7518:
I know Paul but if I'm not able to send a pull request I do not have any other options left.
Im getting always:
image

It seems that only a certain group of people are allowed to even send a pull request. Maybe someone needs to add me somewhere?

mattjgalloway:
@firefox7518 - you're trying to *push* to the emoncms repo. That's not the only way for PRs to work. And infact that's not usual either, in this scenario.

What you need to do is put your work on a branch in *your* repo and then issue a PR from that branch in your repo to the main repo.

Feel free to ping me when it's there and I can help review it.

firefox7518:
Hi Matt

It's actually in a branch! I made a new branch after syncing down the emoncms/dashboard repo and put in my changes there. And then I hit "Pull Request" and added my comments to it. Is there another way of doing it. We are using git enterprise in the office for all our work and there it's working like this without an issue :-/

mattjgalloway:
It should work just fine if you've got the work in a branch on your repo, then you do a PR to the main repo.

I'll check properly later when I'm on my computer but I don't see why you can't do that.

firefox7518:
Thx a lot for helping me here...

What do you think about my suggestion to use a floating box like other tools (Photoshop or Gimp)?

mattjgalloway:
> What do you think about my suggestion to use a floating box like other tools (Photoshop or Gimp)?

I think it's a decent idea. I'm not certain on being able to move it around (assuming that's what you mean by this?). But I do think it should float above the content, so that when you scroll down / across it moves with the screen.

mattjgalloway:
@firefox7518 - on your PR issue. I notice that your work doesn't seem to be on a branch in your repo:

screen shot 2016-09-23 at 17 23 21

As you can see, there's only 1 branch, `master`.

Ideally you would put this commit (looks like you've got just 1 commit with your work) onto a new branch and then issue a PR from that branch back to [emoncms/dashboard](https://github.com/emoncms/dashboard). Usually, when you have some work on a branch in your repo, GitHub will automatically show you a button on the upstream repo page to create a PR from your branch.

firefox7518:
Yes moving it around is a must. Even in Gimp or Photoshop you sometimes have to move it around.
I used a js script which allows you to move it anywhere you want :-) But I never came to the idea to make it a box until I saw your additions here with the undo buttons etc. Then it was obvious to me to use something like this:
image

About the branch thing. It's really funny. GitDesktop clearly shows it to me as a branch:
image
And as you can see it offers me to do a pull request from Widget-enhancements to master. This normally should be ok like this.  Any ideas what could be the issue? I try it again.

mattjgalloway:
Hmm maybe you're working with your remote being emoncms/dashboard rather than your fork? And that branch you see is just a local branch? It does seem like you're trying to push to the emoncms/dashboard repo given the error you're getting.

firefox7518:
Did a first test by changing some stuff in boostrap.min.css and dashboard_edit_view.php. Just a first start.
I used a js script which actually lets you move the box around in the window :-) Positioned on the right:
image

What do you all think?

Paul-Reed:
As a concept, that's really good as it maximizes the workspace  (as Photoshop designers discovered!)
However I'm wondering how the dashboard functions can be displayed aesthetically, and in a manner that users would find intuitive. 

Paul

mattjgalloway:
Nice work! I'll need try it out to see how it feels. Can you PR it maybe? Don't worry - it won't be merged until it's ready. PRs are still a good way to get feedback on direction both of the code and the idea.

A side note - please don't edit `bootstrap.min.css`. This is an external dependency and we shouldn't be editing that. It's the minified version of Bootstrap, the framework from Twitter.

firefox7518:
Created a pull request for this changes. I aligned the buttons as good as it gets via direct style injections in designer.js and dashboard_edit_view.php. This way we can use bootstrap styles to define them. 
Added on file in the dashboard folder which is necessary for the moving part of it. Happy testing with feedback :-) It's of course just a start. We can think of using icons for back and forward, deleting etc. Which makes it look nicer I think.

Andi

chaveiro:
I like this new developments, nice work. 
About font, i personally don't see a reason for it being configurable.
Why not keep a default for all the platform that will also easy out development?

Paul-Reed:
Thanks Matt

Paul

PR #17, opened 2016-09-10
Add apikey to list query, so that you can view dashboards with the readkey

mattjgalloway:
This fixes it so that with the read key passed as the `apikey` parameter in the URL query, you can see a dashboard. Without this fix, widgets are blank because the `list` query doesn't pass the API key, so you don't get any results back.


Paul-Reed:
Thanks Matt. This is really useful (as you already know) because it enables mobile users to pin the URL to their home screen, or share dashboard access without revealing the write api key. :+1: 

Paul

PR #18, opened 2016-09-23
Addition of Widgets

firefox7518:
- created bar2 widget with options to add a title, choose color of title
and font of widget
- created windrose widget (already included since some months)
- created custom text widget to allow to select Font, color and Size,
fits nicely to the new feedvalue_custom. No HTML allowed.
- Created digital compass widget
- created feedvalue_custom which allows for selection of feed, color,
font and size.
- Created Jgauge2 (already in Master since some months)
- Enhanced LED widget to allow a size value to be entered (better
resizing this way)
- feedvaluesmall widget as a smaller copy of the feedvalue widget. Text
size identical to normal-text widget
- added a door widget which allows to visualize simple door open/closed
activity incl. colour selection for closed and open state


firefox7518:
Naming of course is open for discussion. Especially the bar2 widget. The setup now would break existing bar widgets if one would exchange the old one with the new one so I named it bar2 for now. Any suggestions?

Same is true for the feedvalue_custom widget.

Paul-Reed:
Thanks Andi. It would have been much easier to review if the PR's had been per widget, so that they could have been individually tested, reviewed and commented upon. As it is, it's either merge all or reject all.

Paul

firefox7518:
Right, didn't thought of that, sorry. I was really surprised it actually worked this time :-)  It seems that GitDesktop doesn't want to sync back my branches, hence I was not able to do the pull request. 

I will do it again per widget. Can you reject this one?

Paul-Reed:
I'll do it from here.

Paul-Reed:
I'm using [Gitkraken ](https://www.gitkraken.com/)now instead of GitDesktop, I'm finding it much easier to use (and more forgiving if things go wrong!)

Paul

PR #19, opened 2016-09-23
Created bar2 widget

firefox7518:
Created bar2 widget with options to add a title, choose color of title and font of widget


Paul-Reed:
Thanks Andreas, yes I think that this is a big improvement, and would like to see this as a replacement, not as an addition to the original bargraph.

However, there are a few issues which I've noticed.

1) The widget does not remember the Title Colour if you re-enter the widget config.

2) The text size of the widget value has shrunk! Although it resizes with the width of the bargraph, it's still significantly smaller that of the bargraph widget of a similar width.
bar

3) The app doesn't play well with graduation lines enabled, making the text size un-proportionate. Also the default setting for Graduations is 'On' which if deployed without specifying 'Num Graduations' naturally results in a small graph, which could confuse users. Could the default setting be 'No'? (this also exists in bargraph).
bar3


Paul

firefox7518:
HI Paul

Thanks a lot for your feedback. I actually found the issue for not picking up the colour. I have forgot to add my changes on designer.js. There it only checks the normal colour picks but not the new title colour. 
I've added the necessary part to also pickup the title colour form the widget.

I changed the bar chart to show the title below the bar, actually between the bar and the value:
image

Do you think this works better like this?

firefox7518:
About the non graduation one: Do you think the value should have the same colour as the title or shall I left it black? At the moment only the title gets the new colour. I kinda like it but of course know that people maybe want to have all the text in the other colour. What do you guys think?

Paul-Reed:
The title at the bottom seems a better solution, it looks clearer and provides more text width.
IMO the value should be same colour as the title, as it ties them both together, and also allows users to create black graphs (if the value is fixed black, you wouldn't see it!)

Paul

firefox7518:
Ok, I will change it to use also the colour, Will rename it then to text colour and not title colour. Or is there a better expression in english? I'm Swiss so I do not know the perfect terminology sometimes :-)

Paul-Reed:
Label Colour?

firefox7518:
Ok enhanced it to use the colour style for the whole widget:
image

Called it Label colour as per your suggestion. Does that sound and look right?
image

Added also this change to the designer.js file to use lable_colour as key. It's now so generic that we can use it for other widgets too I think. Let me know your feedback.

Paul-Reed:
The screenshots look great!

As I mentioned in my first post above, what is your view about replacing the existing 'bar' widget with yours? I can't honestly see any benefits in having 2 variants of a bargraph widget (bar & bar2), when bar2 can do exactly what 'bar' does, **plus** other enhancements too. If your code changes were applied to 'bar' instead of creating 'bar2', I'm assuming that existing dashboards wouldn't be broken when users update, and we would then have just one very configurable bar widget.

Also, I'm not sure how you initially set up your dashboard repo, but it's not up to date with the current Master branch. If you look at the proposed changes to designer.js on your [last commit](https://github.com/emoncms/dashboard/pull/19/files), you will see that if merged, it would overwrite chunks of work that @mattjgalloway has done over the previous weeks.

Paul

mattjgalloway:
This looks nice! Good work @firefox7518.

But yes, it looks like you've reverted a lot of my work in this, so we shouldn't merge as-is. I'll comment inline on the code to illustrate which hunks need to stay in your commits and which need to go.

Views/js/designer.js
```diff
@@ -50,13 +50,7 @@ var designer = {
...
 
 
-    'snap': function(pos) {
-        if (designer.grid_size > 0) {
-            return Math.round(pos/designer.grid_size)*designer.grid_size;
-        } else {
-            return Math.round(pos);
-        }
-    },
+    'snap': function(pos) {return Math.round(pos/designer.grid_size)*designer.grid_size;},
```
mattjgalloway: This change needs to be removed from your commit.

Views/js/designer.js
```diff
@@ -65,21 +59,12 @@ var designer = {
...
-                    var z_element = $("#"+z);
-                    var box_element = $("#"+box);
-                    // Only set new box if this box is higher than the existing found element
-                    if (z_element.index() > box_element.index()) {
-                        box = z;
-                    }
-                }
+        if (x>designer.boxlist[z]['left']-4 && x<(designer.boxlist[z]['left']+designer.boxlist[z]['width']+4)) {
+            if (y>designer.boxlist[z]['top']-4 && y<(designer.boxlist[z]['top']+designer.boxlist[z]['height']+4)) {
+            box = z;
```
mattjgalloway: This change needs to be removed from your commit.

Views/js/designer.js
```diff
@@ -65,21 +59,12 @@ var designer = {
...
             }
         }
+        }
```
mattjgalloway: This change needs to be removed from your commit.

Views/js/designer.js
```diff
@@ -213,6 +198,10 @@ var designer = {
                 options_html += designer.select_feed(box_options[z],feedlist,0,val);
             }
             
+			else if (options_type && options_type[z] == "feedid2"){
+                options_html += designer.select_feed(box_options[z],feedlist,0,val);
+            }
+
```
mattjgalloway: Where are you using this? I can't see `feedid2` anywhere else in this commit.

Views/js/designer.js
```diff
@@ -355,17 +344,6 @@ var designer = {
...
-    'delete_selected_boxes': function(){
-        if (designer.selected_box) {
-            delete designer.boxlist[designer.selected_box];
-            $("#"+designer.selected_box).remove();
-            designer.selected_box = 0;
-            designer.draw();
-            designer.modified();
-            $("#when-selected").hide();
-        }
-    },
```
mattjgalloway: This change needs to be removed from your commit.

Views/js/designer.js
```diff
@@ -377,74 +355,6 @@ var designer = {
...
-        var targetTagName = e.target.tagName.toLowerCase();
-        if (targetTagName === 'input' || targetTagName === 'textarea') return false;
-
-        if (designer.selected_box) {
-            designer.delete_selected_boxes();
-            return true;
-        }
-        return false;
-    },
-
```
mattjgalloway: This change needs to be removed from your commit.

Views/js/designer.js
```diff
@@ -570,46 +480,26 @@ var designer = {
...
-                    if (designer.handle_delete_key_event(e)) {
-                        e.preventDefault();
-                    }
-                    break;
-                default:
-                    // Key not handled
-                    break;
-            }
-        });
-
```
mattjgalloway: This change needs to be removed from your commit.

Views/js/designer.js
```diff
@@ -570,46 +480,26 @@ var designer = {
...
         // On save click
         $("#options-save").click(function(){
             $(".options").each(function() {
                 if ($(this).attr("id")=="html"){
                     $("#"+designer.selected_box).html($(this).val());
                 }
-                else if ($(this).attr("id").substring(0,6)=="colour"){
+                else if ($(this).attr("id")=="colour"){
```
mattjgalloway: This looks suspicious. The fact it was substring matching before tells me that there's a reason for that. Have you checked other widgets that they all still work? Maybe there are widgets with fields like `colour1` and `colour2`. In that case, this change will break those.

Views/js/designer.js
```diff
@@ -570,46 +480,26 @@ var designer = {
...
                     colour = colour.replace("#","");
                     $("#"+designer.selected_box).attr($(this).attr("id"), colour);
                 }
+				else if ($(this).attr("id")=="title_colour"){
+                    // Since colour values are generally prefixed with "#", and "#" isn't valid in URLs, we strip out the "#".
+                    // It will be replaced by the value-checking in the actual plot function, so this won't cause issues.
+                    var colour = $(this).val();
+                    title_colour = colour.replace("#","");
+                    $("#"+designer.selected_box).attr($(this).attr("id"), title_colour);
+                }
```
mattjgalloway: If you changed your `id` to `colour_title`, and remove the change above which I said is suspicious, then you won't need to do this here as it'll be picked up in the above if-statement.

Views/js/designer.js
```diff
@@ -625,37 +515,21 @@ var designer = {
...
-             designer.delete_selected_boxes();
+            if (designer.selected_box)
+            {
+                delete designer.boxlist[designer.selected_box];
+                $("#"+designer.selected_box).remove();
+                designer.selected_box = 0;
+                designer.draw();
+                designer.modified();
+                $("#when-selected").hide();
+            }
```
mattjgalloway: This change needs to be removed from your commit.

Views/js/designer.js
```diff
@@ -625,37 +515,21 @@ var designer = {
...
-            if (designer.selected_box){
-                var selected_box_element = $("#"+designer.selected_box);
-                var prev_element = selected_box_element.prev();
-                if (prev_element.length > 0) {
-                    selected_box_element.insertBefore(prev_element);
-                    designer.draw();
-                    designer.modified();
-                }
-            }
-        });
```
mattjgalloway: This change needs to be removed from your commit.

widget/bar2/bar2_render.js
```diff
@@ -0,0 +1,352 @@
...
+		}
+	};
+
+	var fontoptions = [
+					[5, "Arial Black"],
+					[4, "Comic Sans MS"],
+					[3, "Courier New"],
+					[2, "arial"],
+					[1, "Georgia"],
+					[0, "Impact"]
```
mattjgalloway: I might suggest adding `sans-serif` in here, and `Helvetica` (can't be without that!).
I might also like a "default" option whereby it doesn't override but picks up from the dashboard stylesheet (i.e. current behaviour). Otherwise we can't really ever use this as the default bar (see @Paul-Reed's comment about merging these changes into the current bar widget) as it won't be backwards compatible.

Paul-Reed:
@mattjgalloway Re my other comment about changing 'bar' instead of adding 'bar2', would it be easier to  create a new fork from dashboard master, and apply changes to the 2 files there?

widget/bar2/bar2_render.js
```diff
@@ -0,0 +1,352 @@
...
+		size = width/2;
+	else
+		size = height/2;
+
+	size = size;
+	
+	
+	if (font == 0)
+	{
+		fontname = "Impact"
```
mattjgalloway: Instead of doing this, can you key into the `fontoptions` object to get the font name?

mattjgalloway:
> @mattjgalloway Re my other comment about changing 'bar' instead of adding 'bar2', would it be easier to create a new fork from dashboard master, and apply changes to the 2 files there?

To be honest, I would start this PR again. I would re-branch off master and apply the work needed to get the original bar widget to have all the goodness that this work brings in.

The reason I say that is that this PR has extra commits in it that will end up polluting the history of the main repo. It brings in a new "digital compass" widget that then gets removed again.

So I would keep it simple and branch again, then force push that to the branch that this PR is from. Or just create a new PR.

Git can be confusing sometimes, and I'm happy to help out if pointers are needed!

Paul-Reed:
My thoughts too Matt.
I know it's causing a lot of work Andreas, but there are too many underlying distractions with this PR, and much safer to start afresh with a clean fork/branch.
I've done the same in the past and I know how disheartening it is after committing so much time & effort, but this is a nice addition and worth the effort.

Paul

firefox7518:
Ok, I close it and start over.

Paul-Reed:
Andreas, this may help - http://stackoverflow.com/questions/7244321/how-do-i-update-a-github-forked-repository

Paul

widget/bar2/bar2_render.js
firefox7518: There I need some help. As said I'm not really a pro when it comes to js coding. 
Do you have a suggestion? I also thought of having a centralized repo for this. How about putting this in render.js/designer.js? Then we could use it for the other widgets too. Do you have a suggestion?

mattjgalloway: It's hokey, but you could make `fontoptions` a global variable, and then use it in `bar2_widgetlist` and also here.

But, I just took a look at how the widgets are loaded, and I think your idea of putting it in `render.js` is a good one. My suggestion would be to have something like this at the bottom of that:

```
function available_fonts() {
  return [
    [6, "sans-serif"],
    [5, "Arial Black"],
    [4, "Comic Sans MS"],
    [3, "Courier New"],
    [2, "Arial"],
    [1, "Georgia"],
    [0, "Impact"]
}
```

And then you can use that instead of making your own `fontoptions` on L44. And you can also use it here to get the font, like so:

```
var fontoptions = available_fonts();
var fontname = ""
$.each(fontoptions, function (i, thisFont){
  if (thisFont[0] == font) {
    fontname = thisFont[1]
    return false;
  }
})
```

(That's untested, completely, but it should be along the right tracks!)

widget/bar2/bar2_render.js
```diff
@@ -0,0 +1,352 @@
...
+		}
+	
+	
+	if (font == 1)
+	{
+		fontname = "Georgia"
```
mattjgalloway: I just noticed you have some lines without an ending `;`. It'd be nice to put them in. It's good to have a consistent style and we have semicolons everywhere so let's keep them.

firefox7518: will check that, thx

PR #22, opened 2016-09-25
Feedvalue_custom widget created

firefox7518:
Features:

- Colour Selection
- Font Selection
- Size Selection
- Unit Position (from Matt Galloway)


mattjgalloway:
Nice!

I think we definitely need to factor out that font handling like I suggested before if we're using it here too.

Also - can we just add these options to the existing `feedvalue` widget please?

PR #23, opened 2016-09-25
Created option to enter size for LED.

firefox7518:
Workaround for resizing issues. Allows to specify a size during creation of widget.

### DIFF
--- a/widget/door/door_render.js
+++ b/widget/door/door_render.js
@@ -1,0 +1,168 @@
+/*
+   All emon_widgets code is released under the GNU General Public License v3.
+   See COPYRIGHT.txt and LICENSE.txt.
+
+    ---------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


mattjgalloway:
This is lovely. 2 comments though:

  1. Can you remove the `door_render.js` from this please? It's unrelated I believe? Maybe another PR for that one?
  2. I'd prefer if the LED scaled to the size of the widget, rather than setting a size. I'm a little concerned with this that it won't be backward compatible because what if you load a dashboard where there's no size set? But if we size it to fit, then it should be fine and dandy. I believe the current way is the LED defaults to 30x30, so 20x20 for the LED and 5 padding on all sides. We could keep the padding at 5 and make the LED fit centrally in what's left?

chaveiro:
I recommend that Led size should scale with the widget size.

PR #24, opened 2016-09-27
Toolbar creation in dashboard edit mode

firefox7518:
Created a photoshop like toolbar to allow for additional buttons and function without the restrains of top navigation row.
- Toolbar widget is movable within the dashboard area.
- Buttons left aligned width at 100% to make them look identical in width

Open issues:
- Button for Visualizations is not the same size even if he has width 100%

To do:
- "not modified", orange "Change" alarm and "Saved" information should go to the bottom of the tool bar to have a less jumpy feeling.


firefox7518:
No, did not change them. Let me start again a pull request with freshly updated files.

mattjgalloway:
Ah, it appears that all the lines have windows line endings on them now, doh! What editor are you using? Perhaps open it in Atom or something and fix the newlines?

PR #30, opened 2016-10-05
Merge undo states if it's the same action multiple times in a row

mattjgalloway:
This change means that if you move something left 5 times, then the undo states are all squashed into one. This is more natural. Similarly if you move things forward or backward multiple times, it's just one undo state.

### DIFF
--- a/Views/js/designer.js
+++ b/Views/js/designer.js
@@ -41,6 +41,7 @@ var designer = {
     'undostack': [],
     'redostack': [],
     'nextundostate': null,
+    'lastundoidentifier': null,
 
     'init': function(){
         designer.cnvs = document.getElementById("can");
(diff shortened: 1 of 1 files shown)


Paul-Reed:
Thanks Matt, only minor issue I can see is that if you start with a 'saved' dashboard, enter edit mode, and move an item, the message 'Changed, press to save' is displayed (as it should!).
But if you then undo the item move, so the item is returned back where it started, the message 'Changed, press to save' continues to be displayed, and clicking it results in a further message 'ERROR - Could not save dashboard. Dashboard not updated'.
Is there a way to refresh the 'press to save function' when using undo, so it recognizes that there aren't any changes.

Paul

mattjgalloway:
Yeh well spotted! This issue actually exists before this, and even before my undo/redo work. If you moved something then moved it back to exactly where it was, the save didn't work. Same error.

I should have filed an issue for this but totally forgot. Care doing that? Or I can do it later. I'm happy to figure out a fix for it. But as I say, orthogonal to this work.

Paul-Reed:
Thanks Matt.

Paul

PR #32, opened 2016-10-09
Fix saving when the content hasn't actually changed

### LINKED ISSUE #31: No changes to save
If you start with a 'saved' dashboard, enter edit mode, and move an item, the message 'Changed, press to save' is displayed (as it should!).
But if you then undo the item move, so the item is returned back where it started, the message 'Changed, press to save' continues to be displayed, and clicking it results in a further message 'ERROR - Could not save dashboard. Dashboard not updated'.

Paul

ndo

mattjgalloway:
See #31. There was a problem where if you saved when the content hadn't actually changed, then there was an error. Of course it's not actually an error, but rather the content hasn't changed, so there's nothing to save. This could happen if you moved something to the left and then back again. With undo/redo now it's more likely to happen, so important to fix. This fixes it.

Fixes #31.

### DIFF
--- a/Views/dashboard_edit_view.php
+++ b/Views/dashboard_edit_view.php
@@ -81,16 +81,29 @@ if (!$dashboard['height']) $dashboard['height'] = 400;
 
     render_widgets_start(); // start widgets refresh
 
+    var lastsavecontent = $("#page").html();
+
     $("#save-dashboard").click(function (){
-        //recalculate the height so the page_height is shrunk to the minimum but still wrapping all components
-        //otherwise a user can drag a component far down then up again and a too high value will be stored to db.
-        designer.page_height = 0;
-        designer.scan();
-        designer.draw();
-        console.log("Dashboard HTML content: " + $("#page").html());
-        var result=dashboard.setcontent(dashid,$("#page").html(),designer.page_height)
-        if (result.success) {
-            $("#save-dashboard").attr('class','btn btn-success').text('<?php echo _("Saved") ?>'); 
+        var currentcontent = $("#page").html();
+
+        var success = false;
+        if (currentcontent === lastsavecontent) {
+            // If it's not changed, just bypass actual saving and assume success
+            success = true;
+        } else {
+            //recalculate the height so the page_height is shrunk to the minimum but still wrapping all components
+            //otherwise a user can drag a component far down then up again and a too high value will be stored to db.
+            designer.page_height = 0;
+            designer.scan();
+            designer.draw();
+            console.log("Dashboard HTML content: " + currentcontent);
+            var result=dashboard.setcontent(dashid,currentcontent,designer.page_height)
+            success = result.success;
+        }
+
+        if (success) {
+            $("#save-dashboard").attr('class','btn btn-success').text('<?php echo _("Saved") ?>');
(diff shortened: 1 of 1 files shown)


Paul-Reed:
Matt, don't know if it's a local caching issue, but;

1) If I move an item I see the 'Changed, press to save' button, if I undo the change without saving, so it's back where it started, the button 'Changed, press to save' is still displayed despite there being nothing to save (although the subsequent error message does not appear, which is a step forward).

2) Move an item to the left and press save (save button now green). But then change your mind and undo the change, returning the item to the right, but the save button is still green, and remains green if you alternate redo/undo.

Do you see the same?

Paul

mattjgalloway:
> 1) If I move an item I see the 'Changed, press to save' button, if I undo the change without saving, so it's back where it started, the button 'Changed, press to save' is still displayed despite there being nothing to save (although the subsequent error message does not appear, which is a step forward).

So I left it like that on purpose. I felt that this was the best approach because that's similar to how we're used to seeing things work. For example if you make some edits in something like Photoshop, but the state is actually the same as it started as, it'll still show a '*' or something to indicate that you've made modifications since you opened it. Does that make sense? WDYT?

> 2) Move an item to the left and press save (save button now green). But then change your mind and undo the change, returning the item to the right, but the save button is still green, and remains green if you alternate redo/undo.

Ooo ouch. I doubt that's anything to do with my changes, but I'll take a look. That sounds, bad. I'll take a look now.

mattjgalloway:
> Ooo ouch. I doubt that's anything to do with my changes, but I'll take a look. That sounds, bad. I'll take a look now.

Gah this is a problem with undo. I'll raise an issue and file a separate PR for it (it's unrelated to this change, but good find!).

Paul-Reed:
> ...it'll still show a '*' or something to indicate that you've made modifications since you opened it. Does that make sense? WDYT?

Intuitively, I would have expected the 'Press to save' button to be active only if there was something to actually save, otherwise it's an unnecessary user action with no end benefit, and raises doubt that there may be other 'overlooked' changes elsewhere on the dashboard.
That's only my opinion, and shouldn't detract from the main purpose of the commit - suppressing the misleading 'error' message, which this fixes.

Merge or re-consider?

Paul

mattjgalloway:
> Intuitively, I would have expected the 'Press to save' button to be active only if there was something to actually save, otherwise it's an unnecessary user action with no end benefit, and raises doubt that there may be other 'overlooked' changes elsewhere on the dashboard.

I thought this at first as well, but then I considered how other apps do it (although, not all do, so it's by no means incorrect to do it the way you suggest). I then realised that I quite like the concept that once the button goes orange, it's orange until you save, even if that save is a no-op. It felt to me like the button going green again is one of those situations where it's technically correct, in that nothing needs saving, but isn't actually intuitive. The user might have done 100 edits and magically the dashboard is back to where they started, but they *think* they've made changes, and the button being green again would feel like a bug. We'd have to introduce a state where the button goes green and says something like "Back to where you started" (ok, facetious, but you get the idea).

Does that make sense at all?

Paul-Reed:
Don't agree Matt, but it's not a big issue - just a difference of opinion!
You've fixed the main problem (many thanks)

Paul

PR #33, opened 2016-10-09
Allow selecting multiple elements at once

mattjgalloway:
This allows you to select multiple elements at the same time, so that you can move them around and bring them forwards, backwards.

Current limitation is that moving around is only supported by using the arrow keys. I want to support moving using the mouse, but it was more complicated. The way it is is still very useful though, IMO.

Would love some feedback on this.

### DIFF
--- a/Views/js/designer.js
+++ b/Views/js/designer.js
@@ -528,58 +528,62 @@ var designer = {
 
         designer.start_save_undo_state();
 
+        var left_shift = 0;
+        var top_shift = 0;
+        var snap_amount = Math.max(designer.grid_size, 1);
+        switch(e.keyCode) {
+            case 37: // Left
+              left_shift = -snap_amount;
+              break;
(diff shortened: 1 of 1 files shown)


Paul-Reed:
Matt, were you aware of this, when multiple elements are dragged off screen;

edge

Paul

mattjgalloway:
Hmm. Yes, now I seem to recall seeing that! (Did this work a while ago and forgot about this problem).

WDYT? On one hand I don't think it's that bad. But it's not ideal either. Let me see if I can fix it quickly.

mattjgalloway:
@Paul-Reed - there, fixed. It's amazing how you see the woods for the trees when you just come back to it a while later!

Paul-Reed:
Great! this is a really nice addition. Moving each element individually and aligning them again was a pain!.
Thanks Matt.

Paul

PR #35, opened 2016-10-09
Make undo and redo set modified state correctly

### LINKED ISSUE #34: Undo doesn't raise modified state.
Steps to reproduce:
1) Make some changes to a dashboard.
2) Save the dashboard.
3) Press undo.

Expected action:
The save button goes orange and you can save the dashboard.

Actual action:
The save button stays green and you can't save the dashboard.

mattjgalloway:
We had an issue where if you made some changes, then saved, then did an undo or redo, you couldn't save until you made another change. This was because undo and redo weren't setting the modified state. This fixes that.

Fixes #34.

### DIFF
--- a/Views/js/designer.js
+++ b/Views/js/designer.js
@@ -113,6 +113,7 @@ var designer = {
         designer.selected_box = 0;
         designer.scan();
         designer.draw();
+        designer.modified();
         designer.check_undo_state();
     },
 
@@ -129,6 +130,7 @@ var designer = {
         designer.selected_box = 0;
         designer.scan();
         designer.draw();
+        designer.modified();
         designer.check_undo_state();
     },
 


Paul-Reed:
Issue still persists here Matt.
Stock dashboard clone & diff applied.

Paul

mattjgalloway:
> Issue still persists here Matt.
> Stock dashboard clone & diff applied.

Does it happen when you do exactly the steps I outline in #34? I'm just wondering if there's another case that I'm missing.

Paul-Reed:
Yes, exactly the same.

Paul

Paul-Reed:
This is what I see Matt;

undo

Paul

mattjgalloway:
Oh wow, that's very strange. I do those exact steps and it doesn't do that. Gah! I'll see if I'm not going mad...

Paul-Reed:
Sorry Matt, re-applied the diff again (after trying https://github.com/emoncms/dashboard/pull/33) and now for some reason it's working as you expected.
I had flushed my browser cache previously, and also checked git status so I know that the diff was correctly applied. So can't explain!!!

Anyway, all good now.

Paul

mattjgalloway:
Great stuff!

PR #37, opened 2016-10-14
Toolbox upgarde

firefox7518:
Added a floating toolbox to dashboard designer which has the option to move the toolbox within the dashboard. Arranged the buttons to fit into the box.


firefox7518:
Of course the style stuff should go into the emon.css file but for discussing and finding the final style I left it like this, its easier for me at the moment.

Paul-Reed:
Thanks Andreas, now merged into 'palette' branch & forum post updated.

PR #41, opened 2016-10-18
Use jQuery instead of normal JS

mattjgalloway:
This changes the palette to use jQuery instead of normal JS. Just a bit nicer and more portable.

It also fixes an issue where if you dragged while clicking on one of the buttons, it moved the palette.


Paul-Reed:
Thanks Matt

Paul

PR #44, opened 2016-10-18
Removing submenu completely by adding floating edit icon

firefox7518:
Testing PB66's suggestion to completely remove the submenu bar. To achieve this I created a floating edit button with opacity to reduce the visibility. Using the Jquery code from Matt Galloway.

### DIFF
--- a/Views/dashboard_menu.php
+++ b/Views/dashboard_menu.php
@@ -13,10 +13,11 @@
 
     $result = "";
 
+/*
     if ($type=="view" && isset($id)) {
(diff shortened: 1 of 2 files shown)


firefox7518:
Naming and maybe icon have to be created/found for better understanding. Touch events missing so only draggable with mouse at the moment.

Paul-Reed:
This looks better than the earlier forum screenshot, especially because of it's opacity.

Thanks Andreas, I'll put some feedback in the forum.

Paul

PR #46, opened 2016-10-20
Changed palette position & transparency in view mode

pb66:
Aligned the view and edit palette positions so the edit/view toggle
button position doesn't move when toggling.
Made the pallete/button transparent  in view mode.
Moved both palettes tighter into the top RH corner to keep the "edit"
icon as far out of the main screen area as possible durnig normal
viewing.


firefox7518:
Hi Paul

We now have to PR floating around with changes on the same. I've reworked the edit button slightly too and used the gear icon instead of the wrench icon with text. Maximizing the usage of the dashboard area.

PR #47, opened 2016-10-20
Edit button reworked and double slash fixed

firefox7518:
Reworked edit button:
- Flat style
- Opacity 0.7 to help shine through stuff underneath
- Invisible DIV around for moving it it still occupies something
- Top right corner

Fixed double slashed when using the view button on the editor.

### DIFF
--- a/Views/dashboard_edit_view.php
+++ b/Views/dashboard_edit_view.php
@@ -41,7 +41,7 @@ if (!$dashboard['height']) $dashboard['height'] = 400;
 <div id="toolbox" class="toolbox" style="background-color:#ddd; padding:10px; position:fixed;z-index:1; border-radius: 15px 15px 15px 15px; border-style:groove; width: 130px; height: auto; top:110px; right: 50px;">
 	<span id="dashboard-config-buttons">
 	<button id="dashboard-config-button" style="float:left"  class='btn' style="float:right" href='#dashConfigModal' role='button' data-toggle='modal'><span class='icon-wrench' title= <?php echo ("Configure dashboard"); ?>></span></button>
-	<a class='btn' style="float:right" href=' <?php echo ($path.'/dashboard/view?id='); ?><?php echo $dashboard['id']; ?>' ><i class='icon-eye-open' title=<?php echo ("View Mode"); ?>></i></a>
+	<a class='btn' style="float:right" href=' <?php echo ($path.'dashboard/view?id='); ?><?php echo $dashboard['id']; ?>' ><i class='icon-eye-open' title=<?php echo ("View Mode"); ?>></i></a>
 	</span>
 	<span id="widget-buttons"></span>
 	<span id="undo-buttons">
--- a/Views/dashboard_edit_view.php
+++ b/Views/dashboard_edit_view.php
@@ -38,7 +38,7 @@ if (!$dashboard['height']) $dashboard['height'] = 400;
     </div>
 </div>
 
-<div id="toolbox" class="toolbox" style="background-color:#ddd; padding:10px; position:fixed;z-index:1; border-radius: 15px 15px 15px 15px; border-style:groove; width: 130px; height: auto; top:110px; right: 50px;">
+<div id="toolbox" class="toolbox" style="background-color:#ddd; padding:10px; position:fixed;z-index:1; border-radius: 15px 15px 15px 15px; border-style:groove; width: 130px; height: auto; top:50px; right: 10px;">
 	<span id="dashboard-config-buttons">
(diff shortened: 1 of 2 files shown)


pb66:
Hi Andreas
I have just tried this latest version and the "edit-mode" button still has a grey background and doesn't "transform" smoothly into the full palette as the positioning differs between the palettes.

Please try my example and read the notes I made at the time on the forum, I have no preference who submits the changes, but I do think the fully transparent backgrounds (plural) are the way forward in view mode. I have done the PR only to show what I was trying to explain on the forum.

Plus while I really appreciate your efforts to retain the ability to drag the palette in view mode, it is unfortunately IMO, of little or no practical use since the "edit-mode" icon/palatte will always appear back in the top RH corner when navigating the different pages, and even when you do drag it, when clicked the new toolbox/palette open in the top RH corner, forgive me if I'm missing something but I see no reason anyone would ever move the single button, the grab area just confuses the one single option to "click for edit-mode". I think Paul has also pointed this out to you, please correct us if you think there is any value in being able to drag the one button.

Paul

Paul-Reed:
Andreas, thanks for this, I've just loaded up the diff, and it's looking much better 👍 

However before merging, can we discuss the points made by @pb66 and myself about the need to have this button 'floating' instead of 'fixed'. Apart from the added code, and serving no obvious practical purpose, being floating instead of fixed may make it difficult to create the responsive CSS changes needed for mobile/tablet viewing later.
Formatting the icon hasn't been easy, but I feel we are getting much closer now, with some good suggestions being made in the forum, and the [PR submitted by pb66](https://github.com/emoncms/dashboard/pull/46), which I believe was kindly submitted as an example/idea for you to consider, and not to be merged (correct if if I'm wrong Paul!).

Paul

firefox7518:
I think we still should follow minimal  UX guidelines. Somehting which starts an action should be visible as a button to press. If no one sees the issue in having the button there fixed, I'm happy with it to make it not draggable anymore, but then it should not be really bigger than this. On my mobile phone it's now just the right size to not be annoying :-) I just wanted to make it moveable if there is a good bunch of mobile users complaining about it and as long as we do not have a mobile/tablet strategy for the dashboards it could upset them more. Of course I still think the win of space by loosing the submenu bar is it worth to have the button visible but we should make a deal to not set it up too big for now. My Proposal is to leave the button at this size. To show the reason maybe showing my dashboard shows it:
image

As you can see I wanted to make the button at least the same design we already use throughout EmonCMS. It looks the same as the other actions buttons. I really think we should follow this way.

I will add a commit to stop it from moving so that you both can test it again.

pb66:
@firefox7518 - Andreas, I'm still don't understanding why you are so insistent on the almost solid button color, other than "guidelines".

IF the button's background is almost solid then I guess you have a pretty good reason to have the button as small as possible, so it can still be clicked on a mobile, without it being annoying and also draggable because it is an obstruction.

Whereas in contrast if the button's background was almost transparent then you could have the button larger without the need to move it, making it less obtrusive, easier to use and easier on the eye. I see no downside at all.

Perhaps it strays slightly from these guidelines you refer to in some way, I cannot be sure but it would still be 
- clearly a button
- the same size and shape as other buttons
- grey background (albeit 0.1 opacity)
- same icon size, colour and opacity
so which guidelines does it break? not that it really matters they are only guidelines and a specific applications needs should always take president.

I think it might be a good idea to try and limit the discussions on here to PR's and bug fixes etc, the styling and general development discussion should remain on the forum as it gets complicated with scattered discussions.

@Paul - Don't worry about pulling in my PR it just complicates things, I had waited until you decided to pull in the previous days PR and didn't know Andreas was sitting on another unsubmitted PR, it doesn't really leave room other PR's.

firefox7518:
I think I answered that with the most important reason in the forum thread. Because it's not working on black dashboards. 

I know a lot of people, including myself, who do use black dashboards. Black icon on Black dashboards simply does not work. Adding opacity to make it "milky" still does not make this situation better on dark dashboards.

I would have to write  a code which checks the dashboard background color and if it is black it has to turn into silver or grey. Still black containers are not easy to detect I guess. I would have to read out the DIV positions and the class and then check the coordinates to switch the color. A lot of code and time when it can be simply avoided by a normal button where we can use bootstrap framework.

to solve this Situation I also thinking of putting the configure Icon into the black main menu bar. This bar is always black and we can use a solid color on it as for the other buttons. We could decide to use a slightly different colour then the other icons to show it better. Or we are putting it into the already existing "Setup" menu as a seperate section and name it "Edit Dashboard". This solves all the issues :-)

PR #49, opened 2016-10-22
Changed icons on palette, set size to 80% and add white contour to icon to edit button

firefox7518:
Added a white contour to the icon which lets it still be visible even if the dashboard background is the same color as the icon.
Set size to 80%

Adding custom icons to the palette to maximize usage of it.


Paul-Reed:
The buttons don't seem to align correctly on the right hand side.
The smaller top 4 buttons are correct, but the 4 dropdown buttons appear too wide, and the right hand sides are touching the palette border.

Also, the starting position of the palette is flush to the right browser frame, it would look better if there was some padding, so the palette appeared to be floating, rather than secured to the browser frame.

Paul

palette

firefox7518:
Sorry, your right. I forgot to add also the designer.js changes. The code for the button size is there.
I tried to align them further to fit better into the palette. I also reduced the padding to make the icons looking bigger in the button.

Looking like this now:
image

firefox7518:
Also added 10px from the right so the palette moves in slightly. Can add more if you like more Paul

Paul-Reed:
Better, but still not aligned here

palette

firefox7518:
Can you check your designer.js file? What size do the buttons have there? As you can see on my screenshot they do allign for me

Paul-Reed:
Now looking really good (I think the last alignment problem was due to caching).
Thanks Andreas 👍 

Paul

PR #52, opened 2016-10-28
Optimized value below the bar

TomRottler:
Th value below the bar was not displayed completely if it is used with units.
I just changed the if statement which did not include neither units nor the negative sign.

Tom

### DIFF
--- a/widget/bar/bar_render.js
+++ b/widget/bar/bar_render.js
@@ -256,8 +256,8 @@ function draw_bar(context,
 
 	if (graduationBool == 1)
 	{
-		if (raw_value > 1000)		// Add additional offset to make alignment work for HUGE numbers
-			half_width += (size*0.20)
+		if (raw_value.length + units_string.length > 4)		// Add additional offset to make alignment work for HUGE numbers
+			half_width += (size*0.25)
 		context.fillText(raw_value+units_string, half_width+(size*0.25), height + (size*0.45));
 	}
 	else


Paul-Reed:
Hi Tom, this does help, however there are a number of other issues with this widget, where the text size does not scale properly with the widget size, especially when using graduations, and these probably need resolving first.
I'm aware that Andreas Messerli did some work on this recently, however it was never completed because priority was given to the development of the dashboard floating palette instead.

@firefox7518 are you able to revisit https://github.com/emoncms/dashboard/pull/19 before we look at this change from Tom?

Paul

firefox7518:
I just created a new version of the bar widget and send a PR (#56). If tested ok, I will also incorporate the changes from Tom. If this is ok for Tom?

Paul-Reed:
Tom, can you retest please with the changes that have been merged into the feature/barwidget-rework branch of the dashboard repo.

Thanks

Paul

TomRottler:
I've tested the /barwidget-rework branch and it is working fine. Without a change of my target configuration the font size was very small, so I had to increase the width of the object a little bit. Of cource the size of the complete object defines the width of the bar and also of the font size. This is not a problem for me, it could just be a compatibility issue for others. 

The graduation text is not exactly in the middle of the line: I modified the line 253 to
context.fillText(unitOffset+units_string, width+(size*0.1), curY+(size*0.05));
That about line 169: size = size; Is it necessary?

Thanks for the good new solution.
TOM

firefox7518:
Ah, thanks Tom, I completely forgot to fix the alignment.  You're right about that.

Can you maybe explain more detailed what you mean with the font size was very small? When graduations are on or off?

About line 169 I have to check. I made a lot of changes when trying to make it backwards compatible with the standard bar widget.

TomRottler:
Hi,
to see the small font size just have a look to 
peweu.ddns.net/emoncms/dashboard/view?id=1

TomRottler:
Hi,
small size: If the bar object is very thin, the size of the font shrinks. I used the bar object to show the temperature as a thermometer. The typical capillar is thin.
I tried to show the sample online but the server is not online until monday.
Nevertheless I hope this small expanation helps to understand.
TOM

firefox7518:
Yes I know what you mean. As it is resizing due to the div size I do not know a common value to use for resizing the font size. 

On my Dev Machine a small bar widget looks like this when using Arian normal as a font:
image


But the good thing is you can now switch to arial black as a font which looks like this:
image

better but still not perfect. the issue I see is that the you have to sacrifice a lot of right space for the descriptive values. When going to big with the size * 0.xx values it will lead to cut off digits:

For example just going up from size *0.20 to size*0.25 will of course enlarge the font but the window is too small to show the value:
image

The higher the values the more odd it looks.

Paul-Reed:
Correct me if I'm wrong Tom, but are you meaning the label text size when
graduations are off?
If so, I've noticed the same.

firefox7518:
I just had some luck when looking on his example as the server was responding:
image

Because of that my assumption was that the tick labels are too small to read?

Paul-Reed:
I guess there's a case for making the labels a fixed size instead of
scaling....

Dashboards could start to look very scrappy, if the labels on different bar
graphs were all different sizes.

Are there any other options/thoughts?

firefox7518:
Making them the same size won't work as you are then practically unable to use a higher number of graduations on smaller sizes. I guess I have to find better values to define the size of the bar when in graduations mode. I tested it and using size*0.2 instead of 0.15 gives a little extra size. 

But this was already an issue with the standard bar widget as soon as you use the graduations. I did not change much there or even at all besides the color adaption.

chaveiro:
Let me share my thoughts.
1 . Use only one font across all the dashboard widgets.
2. Dont let user choose other fonts.
3. Do restrict widgets minimal sizes when required.
4. Like above also restrics aspect ratio of widgets if required.
5. Implement generic font sizing algorith for the used font on emoncms widgets, see for example: http://stackoverflow.com/questions/22943186/html5-canvas-font-size-based-on-canvas-size

TomRottler:
Hi Guys,
sorry for initiating this big discussion. Mainly the widget works.
@firefox7518 Andreas, you wrote " I guess I have to find better values...". What about splitting into two areas, the bar area and the text (graduation)? Then you could calculate it easier. e.g. bar = 30% (rest for graduation). HTML uses it for tables. 
@chaviero: 
1 . Use only one font across all the dashboard widgets. --> very good idea, but I think you will not get a lot of friends. 
2. Dont let user choose other fonts. --> outside the widgeds the fonts should be individual for a better design. Headers, important notes and so one need separate attention, made by the font.
3. Do restrict widgets minimal sizes when required. --> Difficult to define the limits, especially for differnt screens
4. Like above also restrics aspect ratio of widgets if required. --> like above. The user can see the result immediately also without a limit. A limit should only be required for the stability of the js.

firefox7518:
Having one font for everything is the issue we already have. The whole point of changing the bar widgets and also others I'm doing at the moment is that you have the option to choose the font style.  I have a 4K 43" Monitor. The actual font looks ugly and is most of the time too big. Even on my FHD Smartphone the fonts looks way out of place.

With the actual solution you now can also use dark backgrounds like the apps as you now can use a different color for the labels which was not possible before. Being free in designing your dashboards was the reason why I'm using EmonCMS for the thousands of PV inverters visualization of my clients. Such a big winning point other solutions didn't have.

And I also think that the users are mature enough to decide if they want to use other fonts and colors or not. What we have to ensure is that we always use the same default style but I definitely do not want to go back to the older version. I even think we should invest more in functionalities like this. In the end everyone should be able to build similar dashboards like the apps dashboards with the designer of emoncms.

I even think we should introduce the possibility to easily upload images and use them as background. This way you can overlay the infos from EmonCMS on a map of your house or in my case the water treatment system of our town where you could show the process steps and data of the water treatement system. With something like this the possibilites of EmonCMS are endless and you have a tool which can visualize everything from energy production or consumption to water consumption or even weather data of building data like open windows and doors. :-) Just my 2 cents.....

chaveiro:
So the major problem that needs fix is incompatibility of a dashboard accross screens sizes and dpi's.
We fix this with adaptative widgets.
- Not user configurable to a specific screen size.

Fixed font ensures the possibility to implement a  generic logic to size the font to all screen sizes dynamicaly (adaptative). This logic is kind of difficult to implement but to bonus is that you only need to do it once and use a lib on all widgets.

So: 
- Fixed font is a requirement.


Ps: Font color is a usefull feature.

Paul-Reed:
@chaveiro Yes, totally agree.
...any thoughts/help to implement?

Paul

firefox7518:
I agree with the logic on how to resize it but disagree with the fixed font behind it. If you can do it for one font you can do it with all.

I'm not able to build something like this. Best I could do I have done.......

Paul-Reed:
@firefox7518 I've only suggested fixing the size, not the font typeface.
Using different font typefaces is a nice addition with little overhead, yet adds individuality and flair to the dashboards.
I guess the beauty of being open source is that more senior developers to you or I can pitch in and help. (see my last post above).

Paul

chaveiro:
A little goggle and maybe using context.measureText mixed with media breakpoints techniques should do the trick, take a look at this, they have live examples at the threads :
http://stackoverflow.com/questions/20551534/size-to-fit-font-on-a-canvas
https://gist.github.com/joubertnel/870190

Additional readings:
https://www.safaribooksonline.com/library/view/making-isometric-social/9781449309664/ch01.html
http://chimera.labs.oreilly.com/books/1234000001654/ch03.html#filltext_and_stroketext

Paul-Reed:
@TomRottler - Now that much of Andreas's work has been merged, could you please review this PR & https://github.com/emoncms/dashboard/pull/53 

Many Thanks
Paul

TomRottler:
I've tested your released bar_render.js and it works fine. Review is ok.
In the configuration some properties are not translated. Is there anybody who adds the missing widget properties, and translates it?

Paul-Reed:
Hi Tom, regarding the translation (I'm presuming you mean German?), you could try reaching out to Anne @sumpfing to see if she would be kind enough to add the missing translation.

Paul

TomRottler:
Hi Paul, it seems to be that Anne (@sumpfing) is not active at the moment. I think I've understand the multi language procedure of the emoncms. So I could also do these part. German is my native language, in case of my international activities I can handle also Spanish and Italian (just technical issues) and I have also some technical translation guys in the background.
For me it is more a problem to handle the git repository... ;-)

Paul-Reed:
Well if you need any help with git, please let us know.
As this 'issue' was about 'Optimized value below the bar' is it OK to close as we've drifted off-topic, and maybe create a fresh issue about translations.

Paul

Paul-Reed:
Thanks Tom

Paul

PR #53, opened 2016-10-28
Optimized view for negative values

TomRottler:
Additionally to the request "Optimized value below the bar" we should not forget negative values for setup the punctiation.
Using the graduation there is a similar behavior. I prepared a solution to view the complete text. The only weak point is that the scaling could have different formats like -100°C and 100.0°C in one object.

TOM


Paul-Reed:
Thanks Tom, please see my comments in https://github.com/emoncms/dashboard/pull/52 regarding other issues with this widget.

Paul

PR #55, opened 2016-10-29
Palette update

firefox7518:
- Added css file for future style sets for the editor view
- Added editable webkit scrollbars to exchange standard scrollbars (smaller, fresh design)
- Added new smaller iconset from iconsdb
- changed looping code to use icons instead of labels in designer.js


Paul-Reed:
Nice update Andreas!

It looks really good with the new set of icons, and overall very functional. I've been looking through it tonight, and posted my notes below.
Most of the issues are minor, but I've included them for your information and comment.

- The 'Toolbox' heading at the top of the palette not only looks good, but is a useful space to grab and move the palette around. However the 'Toolbox' label which is over 33% of the width is a dead zone that cannot be dragged. Is there a way to make all of the heading draggeable?

deadzone

- The 4 dropdown buttons would benefit from the addition of Tooltips, like the 8 funtion buttons, especially as they are only used in emoncms and not in common usage, unlike the undo/redo/settings etc buttons.

- The curser generally used for link-select is a 'hand-curser' which is used within other emoncms menus, however in the palette drop-down, the curser is a 'text-curser' when hovering over the links, which is intuitively reserved for editing/selecting text objects.

menucurser 
buttoncurser


- Are the images emon-icon-settings.png & emon-icon-save.png used anywhere

- The 4 dropdown icons are sized 232 x 232px (about 5k), yet are only displayed 31 x 31px

Paul

Paul-Reed:
Yes, there's only about 1px white outline to the image, and if image is scaled, the outline will disappear. That's why I didn't include it in my review, but yes, I think we need to look at this again in the future, Especially if we should include the icon at all, or edit the dashboards via the main menu ribbon (as @chaveiro recommended)

Paul

firefox7518:
I'm an idiot, I lost the bookmarks to the icons I've used now from iconsdb. Now tried to search for them and went to hundred sites but did not find them anymore! Are they constantly changing their icons?
I found the Vis and Widget one with the keyword "chart". But now completely different icons show up.
Weird.....

Paul-Reed:
Don't get too concerned about the icons, it's not a big issue, and I/we can sort later...

Paul

firefox7518:
Ok Paul, can you retest again? 

I've added the help text but as written in the commit only a generic help text is possible because of the original looping code.

Also added move and pointer cursor to the hopefully correct positions.

How do you like the scrollbar indicator? To thin? I don't think most people will use it to scroll so I used it as an indicator on how much there is to scroll.

Paul-Reed:
I think it's thin, making it harder to select & pull down on a laptop without a scroll wheel/function, but that's just my opinion...
Well it's 1.47am in Liedertswil, and 12.47am in Yorkshire, so let's pick this up tomorrow...

Regards

Paul

Paul-Reed:
Thanks Andreas.
There are a few whitespace errors reported and a couple of unused icon files as mentioned above, which we can tidy up later. Other than that, it's looking really good now.

Thanks

Paul

firefox7518:
Yes, some icons are unused at the moment but the "save" icon I maybe will introduce later but as they are really small it shouldn't be an issue at the moment.

Yeah funny, when looking on N++ there are NO white spaces. I really wonder why and where they are coming from when uploading the file. I even started to copy/paste the code to try to get rid of them. But there is no impact so we can prettify the code later.

Paul-Reed:
I would prefer to remove the 2 unused files, as these things tend to get overlooked over time and litter the repo.
I've saved the deleted image files though, if you need them Andreas.

Paul

firefox7518:
No issue with removing them. Thanks

PR #56, opened 2016-10-30
Bar Enhancements

firefox7518:
- Allow Title
- Allow Label colour
- Allow font selection
- made the value slightly smaller to make it fit better into the bar when graduations are off.
- should not break existing bar widgets. Tested locally but needs testing from others as well.

### DIFF
--- a/widget/bar/bar_render.js
+++ b/widget/bar/bar_render.js
@@ -263,7 +263,7 @@ if (colour_label.indexOf("#") == -1)			// Fix missing "#" on colour if needed
 			var curY;
 			
 			context.fillStyle = colour_label;
-			context.fillText((static_offset+max_value)+units_string, width+(size*0.1), 10+size*0.15);
+			context.fillText((static_offset+max_value)+units_string, width+(size*0.1), size*0.15);
 			var divisions = Number(graduationQuant)+1;
 			for (var y = 0; y < graduationQuant; y++)
 			{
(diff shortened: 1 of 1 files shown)


firefox7518:
Now you can do something like this: 
image

This reminds me a little bit of LCARS Star Trek design :-) I would love to have a theme like LCARS for EmonCMS :-)

Paul-Reed:
I don't want this feature to delay the merging of the palette work, so can you change your barwidget PR to the **feature/barwidget-rework** branch please, and then we can separate the two, and deal with them individually.

Thanks

Paul

firefox7518:
changed

Paul-Reed:
Hi Andreas, I'm not able to change the font style, it remains the same no matter which style I choose.
For example, in the screenshot below, I've selected and saved Comic Sans MS.

temp

Paul

firefox7518:
Hi Paul

Thanks for the feedback. I think I now fixed it. It seems to work here on my local dev machine:
image
left bar is old bar_render.js created and the right one the new one. It did not change the font of the already existing one so this looks good.

Paul-Reed:
Thanks Andreas, yes working now.

Paul

firefox7518:
Did you see any sign of none working standard bar widgets?

Paul-Reed:
No, I prepared some bar widgets before loading your changes, and they integrated well with no problems for me.

Paul

PR #57, opened 2016-11-05
Fix alignment issues and enlarge graduation size

firefox7518:
- fixed bug where when no title was added with graduations an unnecessary : was shown
- enlarged graduation size to be bigger.

### DIFF
--- a/widget/bar/bar_render.js
+++ b/widget/bar/bar_render.js
@@ -97,20 +97,20 @@ function bar_draw()
 			var id = "can-"+$(this).attr("id");
 			var scale = 1*$(this).attr("scale") || 1;
 			draw_bar(widgetcanvas[id],
-								 0,
-								 0,
-								 $(this).attr("title"),
-								 $(this).attr("font"),
-								 $(this).width(),
-								 $(this).height(),val*scale,
-								 $(this).attr("max"),
-								 $(this).attr("units"),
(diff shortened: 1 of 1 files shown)


Paul-Reed:
Thanks Andreas.

**With graduations;**
The value moves about when more digits are added (see image below - watch when the value drops below 1000). Would it be better if the label & value be left aligned with a bit of left padding?

**Without graduations**
Should the font size scale of the label be increased slightly?

Paul

bar

Paul-Reed:
@firefox7518 Any thoughts about my last comments?

Paul

firefox7518:
Hi Paul

Sorry I was a little bit sidetracked at the moment. You're right the value is moving a little bit as I split the code there to accommodate bigger numbers but forgot to align both to the same values :-) Sorry for that.

Will try to fix it over the weekend but as there is not really much feedback about it it seems that there is not a real need for it so far.

Paul-Reed:
Andreas, there doesn't appear to be much forum feedback for ANY
developments at the moment!
But be assured that your contributions are appreciated.

Paul

Paul-Reed:
@firefox7518 are you able to complete as per your previous comments please.

Many thanks

Paul

firefox7518:
Closing it due to deleted repository on my end. Will create a new one with the fix

PR #61, opened 2016-11-22
Cylinder improvements

mattjgalloway:
This brings in 2 things:

 * Makes the cylinder draw to the height of the element, rather than a fixed height.
 * Uses a gradient to draw the bit between the top and the bottom, rather than a self-made gradient using just 6 steps.


Paul-Reed:
Nice changes Matt, especially as it removes more old code than what is added new!!

PR #62, opened 2016-11-23
Update bar_render.js

firefox7518:
Fixed aligning issues when values goes over 1000. Solves moving raw value and title.

### DIFF
--- a/widget/bar/bar_render.js
+++ b/widget/bar/bar_render.js
@@ -97,20 +97,20 @@ function bar_draw()
 			var id = "can-"+$(this).attr("id");
 			var scale = 1*$(this).attr("scale") || 1;
 			draw_bar(widgetcanvas[id],
-								 0,
-								 0,
-								 $(this).attr("title"),
-								 $(this).attr("font"),
-								 $(this).width(),
-								 $(this).height(),val*scale,
-								 $(this).attr("max"),
-								 $(this).attr("units"),
-								 $(this).attr("colour"),
-								 $(this).attr("colour_label"),
-								 $(this).attr("offset"),
-								 $(this).attr("graduations"),
-								 $(this).attr("gradNumber")
-								 );
+								0,
+								0,
+								$(this).attr("title"),
+								$(this).attr("font"),
+								$(this).width(),
+								$(this).height(),val*scale,
+								$(this).attr("max"),
+								$(this).attr("units"),
+								$(this).attr("colour"),
+								$(this).attr("colour_label"),
+								$(this).attr("offset"),
+								$(this).attr("graduations"),
+								$(this).attr("gradNumber")
+								);
 		}
 	});
 }
@@ -166,7 +166,6 @@ function draw_bar(context,
 		size = width/2;
 	else
 		size = height/2;
-		size = size;
 	
 	if (font == 0){fontname = "Impact"}
 	if (font == 1){fontname = "Georgia"}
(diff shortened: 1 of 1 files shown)


Paul-Reed:
Thanks Andreas.
Can you clarify the comments at lines [276, 278, 290 & 292](https://github.com/emoncms/dashboard/blob/feature/barwidget-rework/widget/bar/bar_render.js#L276-L292) please, they appear a little misleading.

Also should the operators on lines [260 & 264](https://github.com/emoncms/dashboard/blob/feature/barwidget-rework/widget/bar/bar_render.js#L260-L264) be >= instead of > because;
9.99 would display as 9.99 = 3 digits
10 would display as 10.00 = 4 digits
10.1 would display as 10.0 = 3 digits  ..and the same for 99.9, 100.0 & 100

I don't know if you've noticed a peculiar 'rounding' in the bar graph (I think that it's existed a while), for example; the originating feed values are 9, 10 and 11, yet the bar graph displays 9.01, 9.99 & 11.0 ???

bar

Paul

firefox7518:
These comments already existed in the original bar_render.js
I just deactivated the whole line as we do not need it anymore. I check if there is a title available and if yes I align the title and the value below the bar centered (in graduation on mode). Because of this we have plenty of room to show up to 100000W without going out of the borders and we actually do not need a different behavior. I just need to check if there is a title or not. 

If there is no title I make the value slightly bigger and move it up too but still plenty of room for bigger values. I will clean it up later if we have proof it is working as expected.

The operator on line 260&264 are also there since the beginning and I didn't touch them actually :-)

You're right the graphs sometimes have a funny behavior. You can solve this 9.01 or 8.99 by resizing the window?!? I do not know why it does this but it's there since I'm using the emoncms dashboards. 

Andi

Paul-Reed:
I wonder if Math.round(raw_value * 10) / 10
would be better instead of toFixed(1)  ??

Paul-Reed:
@chaveiro Have you any idea why we get this peculiar rounding error (screenshot above)?
I've tried a different approach using code below, but the same error still persists...

        function round(value, precision) {
            var multiplier = Math.pow(10, precision || 0);
            return Math.round(value * multiplier) / multiplier;
        }

        if (raw_value>=100)
        {
                raw_value = round(raw_value);
        }
        else if (raw_value>=10)
        {
                raw_value = round(raw_value, 1);
        }
        else
        {
                raw_value = round(raw_value, 2);
        }

chaveiro:
Hi. Not sure maybe due to rounding a float number. See implementation on dial_render.js : 
  if (isNaN(val))
     val = 0;
  else if (val>=100)
     val = val.toFixed(0);
  else if (val>=10)
     val = val.toFixed(1);
  else if (val<=-100)
     val = val.toFixed(0);
  else if (val<=-10)
     val = val.toFixed(1);
  else
     val = val.toFixed(2);

Paul-Reed:
Isn't that[ the same as what we have](https://github.com/emoncms/dashboard/blob/master/widget/bar/bar_render.js#L243-L254) at the moment?

	if (raw_value>100)
	{
		raw_value = raw_value.toFixed(0);
	}
	else if (raw_value>10)
	{
		raw_value = raw_value.toFixed(1);
	}
	else
	{
		raw_value = raw_value.toFixed(2);
	}

Paul-Reed:
...in fact, the same error persists in the dial widget too;

bar

Issue #63, opened 2016-11-26
Rounding issues for numbers <= 10 in widgets


Paul-Reed:
**Test Setup**
Emoncms low-write 9.7.8 | 2016.11.10 with current 'master' version of dashboard installed.
An input fed via browser API calls, which in turn populates a feed - with no other processes defined.

**Results**
Errors occur when the feed value is equal to or less than 10, for example feeding 1, 9, 10 & 11, results in 1.01, 8.99, 9.99 & 11.0 being displayed in the bar & dial widgets, whilst the 'feedvalue' widget (blue text) displays the same value correctly.
The same error appear when the precision changes at 99, 100, 101.

![value1](https://cloud.githubusercontent.com/assets/973580/20639553/17762600-b3bf-11e6-9a1c-16df739c273b.PNG)
![value9](https://cloud.githubusercontent.com/assets/973580/20639555/1c885e38-b3bf-11e6-9e87-46b9fb760897.PNG)
![value10](https://cloud.githubusercontent.com/assets/973580/20639556/1f9a51d0-b3bf-11e6-95e5-3474686b1988.PNG)
![value11](https://cloud.githubusercontent.com/assets/973580/20639560/3254fd98-b3bf-11e6-90ec-983b11718351.PNG)

Paul-Reed:
@chaveiro tried the fix that you suggested in commit https://github.com/emoncms/dashboard/pull/64 which was changes as follows;
[if ((val * 1).toFixed(2) != (associd[feedid]['value'] * 1).toFixed(2) || redraw == 1)](https://github.com/emoncms/dashboard/blob/master/widget/bar/bar_render.js#L77) to Fixed(3) with partial success...
Although the bar widget now displays the correct values - 9.00, 10.00 & 11.0, the actual values [presented to the 'if' statements](https://github.com/emoncms/dashboard/blob/master/widget/bar/bar_render.js#L243-L254) (to 5 dp) are 9.00059,  9.99922  10.99922.
This then still results in errors, because for example; [else if (raw_value>=10)](https://github.com/emoncms/dashboard/blob/feature/barwidget-rework/widget/bar/bar_render.js#L264) will fail because 'val' is 9.99922 and not 10.
Is this something needing fixing - or something we should just accept ....

chaveiro:
Hi Paul, i try to mitigate all this precision issues with https://github.com/emoncms/dashboard/pull/67

Paul-Reed:
[...]

PR #64, opened 2016-11-26
Change precision rules to maintain digit count

Paul-Reed:
For example, to correct;
9.00 = 3 digits
10.00 = 4 digits
11.0 = 3 digits

### DIFF
--- a/widget/bar/bar_render.js
+++ b/widget/bar/bar_render.js
@@ -257,11 +257,11 @@ if (colour_label.indexOf("#") == -1)			// Fix missing "#" on colour if needed
 		}
 	}
 
-	if (raw_value>100)
+	if (raw_value>=100)
 	{
 		raw_value = raw_value.toFixed(0);
 	}
-	else if (raw_value>10)
+	else if (raw_value>=10)
 	{
 		raw_value = raw_value.toFixed(1);
 	}


chaveiro:
Hi paul, the 'real' reason is on this line :
https://github.com/emoncms/dashboard/blob/master/widget/bar/bar_render.js#L77
https://github.com/emoncms/dashboard/blob/master/widget/dial/dial_render.js#L81

// ONLY UPDATE ON CHANGE
if ((val * 1).toFixed(2) != (associd[feedid]['value'] * 1).toFixed(2) || redraw == 1)

here, we need toFixed(3) as it's one step greater than what we use later.
(untested)

Paul-Reed:
Hi @chaveiro 
This commit wasn't to fix the issue described in https://github.com/emoncms/dashboard/issues/63, it was to ensure that the displayed precision was consistent across the number range, ie 10 should be displayed as 10.0 and not 10.00, and 100 and not 100.0. This is already implemented in other widgets [such as the feedvalue widget](https://github.com/emoncms/dashboard/blob/master/widget/feedvalue/feedvalue_render.js#L91-L100) and which does display correctly with these changes.

Regarding the line [if ((val * 1).toFixed(2) != (associd[feedid]['value'] * 1).toFixed(2) || redraw == 1](https://github.com/emoncms/dashboard/blob/master/widget/bar/bar_render.js#L77)
I can't try your suggestion until this evening, however I can't see how that change would affect issue https://github.com/emoncms/dashboard/issues/63. Doesn't it just check if there's been a change in the feedvalue and cause a redraw of the value in the browser?

chaveiro:
Hi Paul, ok thanks for the clarification on this PR.

For #63 I think the reason the value is wrong at the graph is because it stops redrawing sooner due to the rounding on that lines

PR #66, opened 2016-12-08
Code refactoring

chaveiro:
bar_render and dial_render


Paul-Reed:
Thanks @chaveiro , as I've written in issue #63, this doesn't resolve the innacuracy issue, but at least masks it from view.

Paul

PR #69, opened 2017-01-10
Several updates in CURL Widget

RDCH106:
Several updates in CURL Widget

- Added GET/POST method selection
- Added HTTPS option
- Added Timeout option --> Not very useful for now, but I am thinking in the use callbacks
- Pure JSON detection in message payload 
   + If JSON format is detected --> payload is created following this format `{"data":"myJSONdata"}`. For example myJSONdata can be convert as JSON in Node-Red using JSON converter:
image

   + In other cases the data is appended to the main URL

Added new fields translations and updated missing translations.


Paul-Reed:
Hi Rubén, it would be great if this was 2 separate PR's, one for the Translation work which could probably be merged straight away, and a second for the 'Curl' additions.
In respect of the 'Curl' additions, the PR has been flagged up by Codacy as possibly needing a further review;

fail

In more detail, this is the Codacy report;

codacy

Codacy doesn't always get it right! but I would be grateful if you could take a look.

Paul

RDCH106:
@Paul-Reed I start to review the checks. In respect of the translation, the 2 PR come together because the translation are for the updates of the CURL Widget. During the process I have included some missing translations, that's all.

RDCH106:
@Paul-Reed Ready!!

6 issues solved, the last one is very typical in javascript. It recomend `===` because it is a comparation without type conversion (more fast and less resources). But in this case I prefer use `==` using type conversion to ensure that all it is ok.

The `console.log` is for more easy debuging and it helps when there are problems with CORS (Cross Origin Resource Sharing)

Paul-Reed:
Thanks Rubén

Paul-Reed:
If you have the time, it would be great if you could add a post in the forum to inform members of your latest additions.

Paul

glynhudson:
Nice work, thanks!

- sent from my mobile device

PR #71, opened 2017-01-14
Fix button for IE and Firefox

firefox7518:
Changed to onclick solution for the button fixing the not working state on IE and Firefox.
Tested on IE and Firefox and Edge and on EmonCMS in root and subfolder installations.


Paul-Reed:
Thanks Andreas for a speedy fix.

Paul

PR #76, opened 2017-03-17
Added Battery Widget

vkylamba:
Updated.
image

### DIFF
--- a/widget/battery/battery_render.js
+++ b/widget/battery/battery_render.js
@@ -1,0 +1,134 @@
+/*
+   All emon_widgets code is released under the GNU General Public License v3.
+   See COPYRIGHT.txt and LICENSE.txt.
+
+    ---------------------------------------------------------------------
+    Part of the OpenEnergyMonitor project:
+    http://openenergymonitor.org
+
+    Author: Vikas Lamba: vikas13jun@gmail.com
+    If you have any questions please get in touch, try the forums here:
+    http://openenergymonitor.org/emon/forum
+ */
+
+// Convenience function for shoving things into the widget object
+// I'm not sure about calling optionKey "optionKey", but I don't want to just use "options" (because that's what this whole function returns), and it's confusing enough as it is.
+function addOption(widget, optionKey, optionType, optionName, optionHint, optionData){
+  widget["options"    ].push(optionKey);
+  widget["optionstype"].push(optionType);
+  widget["optionsname"].push(optionName);
+  widget["optionshint"].push(optionHint);
+  widget["optionsdata"].push(optionData);
+}
+
+function battery_widgetlist(){
+  var widgets =
+  {
+    "battery":
+    {
+      "offsetx":-80,"offsety":-80,"width":160,"height":160,
+      "menu":"Widgets",
+      "options":    [],
+      "optionstype":[],
+      "optionsname":[],
+      "optionshint":[],
+      "optionsdata":[]
+
+    }
+  };
+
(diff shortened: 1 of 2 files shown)


Paul-Reed:
Thanks for the PR, however it doesn't seem to work here as intended. For example; The widget is configured like this;

settings

and the feed input is 151, I get;

battery

Also, I see that the text size does not scale with the size of the widget, or is user definable - font, label colour (see the bar widget - also in the screenshot above)
It would be good to have some formatting consistency across the widgets.

Paul

vkylamba:
@Paul-Reed Thanks for the review. I have updated the PR. Please have a look.

Paul-Reed:
That's better, it displays the animation OK now.

1. Do we need a widget title? we could of course have several battery widgets displayed on the same page (such as banks of batteries) but with no title, it may not be easy to identify which widget relates to which battery.
2. I think variable rounding of the displayed value would be useful, such as https://github.com/emoncms/dashboard/blob/master/widget/bar/bar_render.js#L252-L258 to enable different precision depending upon the value size. If you were monitoring 999V is there any need to display precision to 0.1V. Similarly for a 1V battery it would be useful to display the value to greater precision.
3. The text value does not appear to be centre aligned to the widget - see screenshot below using Arial Back typeface.
4. Also, for consistency, could the font size be increased to match that of the bar widget? In the screenshot, both widgets are the same size, yet the battery text size is smaller. (so that we have some dashboard consistency)

Paul

align

vkylamba:
1. Added widget title. 
2. Corrected the rounding off. 
3. Text centre aligned.
4. Fixed font size.
image

widget/battery/battery_render.js
```diff
@@ -0,0 +1,183 @@
...
+
+  var fontoptions = [
+    [9, "Arial Black"],
+    [8, "arial"],
+    [7, "Arial Narrow"],
+    [6, "sans-serif"],
+    [5, "Helvetica"],
+    [4, "Comic Sans MS"],
+    [3, "Courier New"],
+    [2, "arial"],
```
Paul-Reed: arial is already listed on line 40

widget/battery/battery_render.js
```diff
@@ -0,0 +1,183 @@
...
+
+      if (font == 0){fontname = "Impact"}
+      if (font == 1){fontname = "Georgia"}
+      if (font == 2){fontname = "arial"}
+      if (font == 3){fontname = "Courier New"}
+      if (font == 4){fontname = "Comic Sans MS"}
+      if (font == 5){fontname = "Helvetica"}
+      if (font == 6){fontname = "sans-serif"}
+      if (font == 7){fontname = "Arial Narrow"}
+      if (font == 8){fontname = "Arial"}
```
Paul-Reed: Arial is already listed on line 95

Paul-Reed:
Thanks for your patience Vikas, the battery widget looks good and matches the existing widgets nicely.

align

Paul

Paul-Reed:
Vikas, I've made a forum entry about the widget https://community.openenergymonitor.org/t/battery-widget-now-added-to-dashboard/3473
I can't find your ID in the forum, are you a forum member? It would be good to introduce yourself in the thread.

Thanks

Paul

vkylamba:
Thank you Paul for merging the PR. I wasn't a member till now. Just signed up.

glynhudson:
Nice work, looks good. Thanks guys!

- sent from my mobile device

PR #77, opened 2017-03-17
Solar widget

vkylamba:
image


Paul-Reed:
Before reviewing this contribution or wasting Vikas's valuable time, it would be useful to discuss with others - @TrystanLea @glynhudson to decide whether this widget (subject to testing) **should** be added to the core dashboard widget list or not.

My own view is that a similar visual effect can easily be created using existing widgets - for example; add an image to a dashboard by adding html to a 'paragraph' widget, and place a 'feedvalue' widget on the next layer.

sun

As it would be added to a increasing list of core dashboard widgets, we need to ensure that  the widget list does not become bloated, or overwhelming for users. 

It's unfortunate that we can't add such widgets individually, and create a community library, similar to what node-red has done with node-red-contrib-nodes.

Paul

glynhudson:
I don't have a problem with adding this widget; sure, you could acheive a similar result using an image. However, I'm sure a js vector renter would be much faster and repeatable.

Paul-Reed:
Hi Vikas, there is a git conflict with your PR, can you pull the remote master branch, and then resubmit another PR please.
I believe the conflict is due to the changes to dashboard_langjs.php which was changed when the battery widget was merged.

Thanks
Paul

Paul-Reed:
Just one very minor issue I've noticed, that pixels of the background colour are showing through between the white outline to the semicircle, and the sunrays.
Is there anyway to minimise this.

Paul

semi

vkylamba:
@Paul-Reed I have fixed it. Please take a look. :)

Paul-Reed:
That's sorted the gaps in the image, thanks.

Could I have your view on another issue please.
The positioning of the text label and feed value appears to change depending upon the size of the widget frame. For example, if the widget frame is adjusted to fit the sunrise, it displays fine;

But if the widget frame is not adjusted, the text label and feed value are displaced;

Printing the text label above the sunrise also creates a further problem, because if the feed value changes, the text label becomes malformed and unreadable.

Should the text label/feed value positioning be relative to the height of the vector graphic instead of the height of the widget frame, so their positioning scales with the graphic?

Paul

Paul-Reed:
Thanks Vikas, I'll merge and fix the fontoptions issue that I mentioned above.

Paul

vkylamba:
Thanks Paul, My latest PR should have fixed the issue. Please go ahead with fixing if its not fixed.

Paul-Reed:
@vkylamba It was a typo on the fontoptions which I've fixed with https://github.com/emoncms/dashboard/commit/6b0073d3cda74f75b7b6cb985e70c93169c5bf39
It's now merged into the the Master branch.

Thank you

Paul

PR #78, opened 2017-04-21
Signal Widget

vkylamba:
image


Paul-Reed:
Hi Vykas, just a few points;

1. If widget config is re-opened again, the app forgets the color assigned to 'Color Signal' it defaults back to black.
2. The app never displays 100% (5 bars), if feed is set to Max Value, it still only displays 4 bars.
3. Is it intended that the app displays the feed value? The app config asks for the 'Units to display' but nothing is displayed.

Paul

wifi

glynhudson:
Nice idea, I like it. 

Could you take a look at the minor issues flaged up by codacy https://www.codacy.com/app/emoncms/dashboard/pullRequest?prid=617420

Paul-Reed:
Any particular errors Glyn?
Most of the errors reported by Codacy for this PR appear to be inherent to the code structure of dashboard, and are flagged up with every widget in the repo.
IMO to resolve most of them would mean a lot of changes to the dashboard module as a whole.

glynhudson:
Yes, agree most if not all are not too important. I'm not well versed in javascript, I don't know if using `==` instead of `===` makes any difference?

vkylamba:
Hi @Paul-Reed
I have fixed the issues, please take a look.
**If widget config is re-opened again, the app forgets the color assigned to 'Color Signal' it defaults back to black.** -- Fixed
**The app never displays 100% (5 bars), if feed is set to Max Value, it still only displays 4 bars.** --Fixed
**Is it intended that the app displays the feed value? The app config asks for the 'Units to display' but nothing is displayed.** -- The widget was not supposed to dispaly the feed value. But I had forgotten to remove the option from the config. I have corrected it now.

Vikas

Paul-Reed:
@vkylamba That is much better, and now it works fine.
There is just one small issue that I can see, where the label text appears to break up at the bottom, against the dashboard background.
Should the title text be on a different layer to reduce interaction with it's background?
typeface2
typeface

vkylamba:
Hi @Paul-Reed,
The title becomes weird when we change the title text. But on page refresh it goes away, and never occurs again. The multilayer context requires adding extra property "z-index" to the canvas. Which I tried with the current code. But it did not work. I don't know if the current structure is designed to recognise this property. 
screen shot 2017-04-23 at 4 41 20 pm

Then when the widget is loaded in the browser this property is not there. I can directly manipulate DOM inside the code to make it work, but that won't be a nice way.

Paul-Reed:
Thanks Vykas, I'm away now until tomorrow evening, so I'll get back to you late.

Paul-Reed:
Thanks Vykas, now merged, although I'm still seeing the artifacts on the label text as per above, even after browser refreshes.
@firefox7518 - I may be mistaken, but did you experience a similar problem with one of your dashboard contributions?

Paul

PR #79, opened 2017-05-03
Min value on battery widget

roy13:
I have applied these to my 9.7 so wanted others to confirm they work in the latest.
Gives minimum to battery for lead acid and lithium characteristics.


Paul-Reed:
Many thanks Roy, the battery minimum value is a nice addition, also the additional themes for the dial.

Paul

PR #84, opened 2017-05-24
Show dashboard edit icon only when logged in

firefox7518:
Dashboard edit icon only visible when logged in


Paul-Reed:
A nice solution!
Thanks Andreas.

Paul

PR #88, opened 2017-06-10
Add fonts properties and unit position to dewpoint

thib66:


Paul-Reed:
Thanks Aymeric, although too late now, can I suggest a better was of developing using git;
If you start with a clean fork of the emoncms/dashboard and before making any changes to it, create a branch of it for example 'humidex'. Now make all your humidex changes in that branch, and submit your PR from that branch.
You can also create another branch from your local fork - 'dewpoint' and work on that branch too, before submitting a PR.
Doing it this way will ensure that each PR will only contain changes relating to that piece of work.

If you look at this PR, you will see that it changes not just the dewpoint file,  but also  the humidex and  frost point files too, so if we had a problem with the dewpoint code at some stage, and we reverted this commit, it would not only remove the dewpoint changes, but also remove humidex and frost point too.

Hope this helps

Paul

Paul-Reed:
Aymeric, thanks again for submitting, the code quality is significantly improved and the two new widgets will be appreciated by the emoncms community.

All 3 file changes are included in this merge, so you can now safely delete your local emoncms/dashboard fork.

I have also created an emoncms forum post about these changes - [see this post](https://community.openenergymonitor.org/t/two-new-weather-widgets-added-to-dashboard/4260)
however, I can't find your id as a forum member to link it to you. It would be nice to say hello in that thread if you get time.

I've noticed that there is no licence attached to your new files, you may wish to add them [using this example](https://github.com/emoncms/dashboard/blob/master/widget/curl/curl_render.js) for consistency (along with your name as author!). 

Kind Regards

Paul

PR #90, opened 2017-06-11
Add an option for the type

thib66:
As I am using a BMP 180 sensor with Raspberry Pi and emoncms I have added a type that emulates a barometer in order to display a pressure


Paul-Reed:
Thanks Aymeric

PR #93, opened 2017-06-13
Decimals widgets

thib66:
changes for more consistency with other widgets like dial widget

### DIFF
--- a/widget/dewpoint/dewpoint_render.js
+++ b/widget/dewpoint/dewpoint_render.js
@@ -211,6 +211,7 @@ function draw_dewpoint(context,
 					else {
 						val = val.toFixed(2);
 						}
+				    val = parseFloat(val);
 				}
 			else 
 				{
--- a/widget/frostpoint/frostpoint_render.js
+++ b/widget/frostpoint/frostpoint_render.js
@@ -206,6 +206,7 @@ function draw_frostpoint(context,
 					else {
 						val = val.toFixed(2);
 						}
+				 val = parseFloat(val);
 				}
 			else 
 				{
--- a/widget/feedvalue/feedvalue_render.js
+++ b/widget/feedvalue/feedvalue_render.js
@@ -200,6 +200,7 @@ function draw_feedvalue(context,
 					else {
 						val = val.toFixed(2);
 						}
+					val = parseFloat(val);
 				}
 			else 
 				{
--- a/widget/humidex/humidex_render.js
+++ b/widget/humidex/humidex_render.js
@@ -201,6 +201,7 @@ function draw_humidex(context,
 					else {
 						val = val.toFixed(2);
 						}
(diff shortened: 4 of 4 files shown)


Paul-Reed:
Aymeric, Out of interest. was this in response  to changing the comparison operators in the 'feedvalue' widget  from == to === (that it did not work because it was comparing a number and a string - not equal types).

Paul

widget/dewpoint/dewpoint_render.js
```diff
@@ -211,6 +211,7 @@ function draw_dewpoint(context,
 					else {
 						val = val.toFixed(2);
 						}
+				    val = parseFloat(val);
```
Paul-Reed: Mixture of tabs & spaces in indent

widget/frostpoint/frostpoint_render.js
```diff
@@ -206,6 +206,7 @@ function draw_frostpoint(context,
 					else {
 						val = val.toFixed(2);
 						}
+				 val = parseFloat(val);
```
Paul-Reed: Mixture of tabs & spaces in indent

thib66:
Hello, I got some issues with decimals but I am unable to reproduce the problem.
My apologies,
Aymeric

Paul-Reed:
No apologies needed, I think that these changes should be made to ensure that they are equal types (numbers and not strings). Do you agree @TrystanLea or @glynhudson ?

Paul-Reed:
Aymeric, I've been looking back through where I had experienced this issue previously, and cannot re-create now. Shall we close this now as there has been no further comments made, and we can always revisit this later if we get the same problems again.

Paul

thib66:
Hello Paul,
Yes you can close the PR.
Thanks,
Aymeric

thib66:
Hello 
@Paul-Reed, I know now why parseFloat is used.
Of course the val variable is in numeric format so it seems not necessary to use parseFloat(). 
But look at the behaviour of the parsefloat function below : 
if val = 9.51 then parseFloat(val) returns 9.51 (no mystery)
if now val = 9.50 then parseFloat(val) returns 9.5
In fact parseFloat() function is used to delete non significant digits (in other terms last zeros in decimals are suppressed)
The parseFloat() is actually used with dial widget for the decimals option set to "Automatic".
I think that whe have to apply this function for all the widgets where we have the choice to select the number of decimals.
Regards,
Aymeric

Paul-Reed:
Aymeric, I'm out of the country, with no system access, so hopefully one of the other emoncms team members will pick this up.
Otherwise, I'm back on the 29th June.

Regards & thanks for your support.

Paul

Paul-Reed:
Yes, agree Aymeric. It makes sense to have consistency across the widgets, and remove the non-significant digits when decimals are set as automatic. Good spot.

PR #94, opened 2017-06-14
fix issue and Fahrenheit conversion deleted

thib66:
Add double quotes in === comparison otherwise Fahrenheit to celsius conversion is ignored.
Celsius to Farhenheit conversion supressed because the Humidex is based from a temperature in °C and the scale is about 29-60


Paul-Reed:
The Fahrenheit calculation are still not correct - for example; Humidity = 72% and temp = 90 degF.
- the result is 46.0

Paul

Paul-Reed:
Yes, working ok now!

PR #95, opened 2017-06-15
heat index widget

thib66:
compute the american heat index from relative humidity and temperature


Paul-Reed:
Aymeric, this widget is not showing up in the dashboard widget list. Tried a reboot and flushed browser cache.

Paul

heatindex

thib66:
Hello, I've just tried 77ccd8c
and the widget is there:
sans titre 2
browser used : firefox 54.0

thib66:
Paul, you should see the widget now... ("_render" was missing in the name)

Paul-Reed:
That's better!
Do you think that there should be a conditional statement to return the original feed value if it's < 26.7 degC, or if >= 26.7 do the heatindex calculations?
Currently, if the temp is less than the heatindex threshold (26.7C / 80F), the calculations return unexpected results, particularly at low temperatures, because the calculations are not intended to be used below the heatindex threshold.
For example at 10.00 degC  & 60% humidity, it returns 8.77 degC instead of 10.00.

Paul

thib66:
Paul,
please have a look at this : 
http://www.nkhome.com/news/when-the-heat-index-is-lower-than-the-actual-temperature/

"Looking at the official chart you notice what seems to be an anomaly-the heat index may be lower than the current temperature. I noticed it this morning on the Kestrel meter I use at my desk. (See image on the right.)

_**The reason for this is that the index assumes a small amount of wind, and also assumes the person is walking. Note, the amount of wind and the walking speed (defined as “Activity” in the list above) is a constant value throughout the table. So in many low-humidity situations (under 30% humidity) with an air temperature of less than 98 F, the heat index will always be lower than the actual air temperature, as explained above."**_

thib66:
Paul,
here is an another exemple : 
http://www.chicagotribune.com/news/weather/ct-wea-0828-asktom-20150827-column.html
Now if you want that heat index = temp when temp <80°F I can do that,
regards,
Aymeric

Paul-Reed:
Yes, but in all of those examples, the actual temperature was far in excess of the the heatindex threshold.
It's understandable that at higher temperatures and low humidity, that the human body can dissipate heat faster, and therefore feel cooler.
But this calculation is not intended to calculate the chill factor, and that's why they have 80 degF threshold.

Check the heatindex tables;
http://thevane.gawker.com/this-is-why-the-heat-index-is-so-important-1609195413
https://www.weather.gov/ama/heatindex
https://en.wikipedia.org/wiki/Heat_index

thib66:
Paul, according to this link : 
http://www.nws.noaa.gov/ndfd/definitions.htm
APPARENT TEMPERATURE is the perceived temperature in degrees Fahrenheit derived from either a combination of temperature and wind (Wind Chill) or temperature and humidity (Heat Index) for the indicated hour. When the temperature at a particular grid point falls to 50°F or less, wind chill will be used for that point for the Apparent Temperature. **_When the temperature at a grid point rises above 80°F, the heat index will be used for Apparent Temperature.  Between 51 and 80°F, the Apparent Temperature will be the ambient air temperature._** 
So in my formula, if I say for ambient temperature > 80°F, I use heat index formula and for ambient temperature <80°F I use this ambient temperature, then my formula is reflecting in fact a partial definition of apparent temperature (as I do not process ambiant temperature less than 50°C with wind chill formula).
So at this step, my formula is neither an apparent temperature nor an heat index...
Regards,
Aymeric

Paul-Reed:
Yes, agree.
So what do we call it to avoid confusion?

Paul

thib66:
Hello Paul,
I propose to keep the name of humidex for the widget and add an option via a pulldown list with 2 choices : 
- strict humidex calculation for the whole temperature range
- humidex only for T>80°F, T used for the rest of the temperature range
Depending on the context, everyone would be free to use the most appropriate formula.
Regards,
Aymeric

Paul-Reed:
Sorry to keep this discussion going... but in the link you provided - http://www.nws.noaa.gov/ndfd/definitions.htm above, NWS are using two separate formulas to determine the apparent temperature;

1) Heat Index derived from Temperature & Humidity
2) Chill factor derived from Temperature and Windspeed

In your formula, you are not able to determine the Chill factor, because you do not know the wind speed, therefore it's not the same.
As far as I am aware, the chill factor cannot be calculated from just the temperature & humidity, because it's the flow of air over the skin that removes heat from the body, and the windspeed cannot simply be replaced by a notional number.

Paul

thib66:
Paul,
I do agree with you, that is why the script I developped cannot be called apparent temperature has the notion of chill factor is not present in the formula for T<50°C. 
the widget could still be called humidex in order to help the choice of the user and with the ability to choose the strict humidex calculation or a kind of apparent temperature only significant for T > 50°C.

Also, I can create a full apparent temperature widget taking in consideration speed, temperature and hymidity feeds.
Unfortunately I would no have the abilty to test in live the result as I have only a BMP 180 pressure sensor and an emonTH ;)

thib66:
Paul, oups! please read 50°F instead of 50°C.
Regards,
Aymeric

Paul-Reed:
> ...or a kind of apparent temperature only significant for T > 50°F

I think that is probably the best option, but what would be displayed when T > 50°F 
I suppose there are 2 options;
1) display the actual temperature or
2) display --°F

Paul

Paul-Reed:
Thanks Aymeric.
Just tried, and it appears to work ok for Fahrenheit, but when using Celsius, a temperature of 9°C, and  rule 2 set to 'display nothing', instead of displaying --°C it displays -295.56°C.
Are you seeing the same?

Paul

thib66:
Hello,
this should be fixed.
Regards,
Aymeric

Paul-Reed:
Thanks Aymeric, this is a nice addition to the dashboard.

As with your previous submissions, I've created a [forum post](https://community.openenergymonitor.org/t/new-weather-widget-added-to-dashboard/4340) to introduce the new widget. Please feel free to contribute to the thread if you wish.

Paul

PR #96, opened 2017-06-16
update of cylinder widget

thib66:
optimization of the code
ability to set the number of decimals
choice of unit position (Back/Front/No display)
the widget now accepts temperature feeds in °F - conversion °F->°C is made in order to benefit from the correct temperature gradiant


Paul-Reed:
Thank you Aymeric.
[Forum post](https://community.openenergymonitor.org/t/cylinder-widget-update/4341) added.

Paul

PR #98, opened 2017-06-19
Isactive widget

thib66:


thib66:
The goal of this widget is to see if a feed is active or not.
This widget could potentially solve the issue explained here : 
https://community.openenergymonitor.org/t/suggestion-widget-if-a-feed-is-old/610
The threshold 1 should not be lower than 5 s as the widgets are refreshed every 5 seconds.
below_threshold1
Below the threshold 1 with default color and default shape
between-threshold1-and-threshold2
Between the threshold and the threshold 2 with default color and default shape
above-threshold2
Above the threshold 2 with default color and default shape
no-network
When there is a network issue (no internet for example) an exclamation mark is displayed
Different shapes are available : circle, square, triagle, star with 5 or 6 spikes

thib66:
dialog

Paul-Reed:
Sorry, but this breaks my dashboard. Browser shows following errors;
error

thib66:
Hello Paul,
probably I miss to add something from my original file.
I will check

thib66:
Paul,
this should work now ...
Regards,
Aymeric

Paul-Reed:
Yes, that was the problem!
Do you think that Colour1/2/3 should be reversed?
For example;

Colour1 = Colour for range below Threshold1
Colour 2 = Colour for range between Threshold1 and Threshold2
Colour 3 = Colour for range above Threshold2

That would then represent the sequence of a feed which was failing.

threshold

thib66:
Hello Paul,
the changes have been applied in order to reverse the colors,
Aymeric

Paul-Reed:
Thanks Aymeric. This has cropped up a number of times in the forum and the way which you've put this together makes it easy for users to utilize it in different ways. For example layering the value on the widget;

isactive

Paul

widget/isactivefeed/isactivefeed_render.js
```diff
@@ -0,0 +1,206 @@
...
+    widgetcanvas[id].fillText("!", (offsetx+dimension/2)*0.85, offsety+dimension*7/8);
+
+  });
+}
+
+function isactivefeed_draw() {
+  $(".isactivefeed").each(function(index)
+  {
+    var feedid = $(this).attr("feedid");
+    var limit1 = $(this).attr("threshold1")*1 || 60;
```
Paul-Reed: Aymeric, just noticed that these default values are not activated when adding a new widget. Neither threshold values or widget colours.
Not a major problem though.

Paul

PR #99, opened 2017-06-21
Isactivefeed optimization

thib66:
dashboard_edit_view.php and dashboard_view.php modified in order to go back to the previous version
isactivefeed_widget.php added in order to respect emonCMS architecture

### DIFF
--- a/Views/dashboard_edit_view.php
+++ b/Views/dashboard_edit_view.php
@@ -113,8 +113,6 @@ function toolboxMove(e) {
 
 <script type="text/javascript" src="<?php echo $path; ?>Modules/dashboard/Views/js/designer.js"></script>
 <script type="application/javascript">
-    var requestTime = Date.now() /1000;
-    var offsetofTime = 0;
     var dashid = <?php echo $dashboard['id']; ?>;
     var path = "<?php echo $path; ?>";
     var apikey = "";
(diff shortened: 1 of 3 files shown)


thib66:
Hello,
@chaveiro , the problem of code polluting existing files should be fixed with the  last changes.
For render.js, in think there is no way to place the added code outside of the script.
Regards,
Aymeric

Paul-Reed:
Could you also take a look at the default values (such as colour) that should be applied when none are specified please. 
Currently the default appears to be black, and not what you have coded.

Paul

thib66:
Hello Paul, 
In fact the default colors coded are just defined to create the object in case initially there are void values (just a security). Actually I don't know how to pass default colors to pulldown lists. Is there somewhere a documentation about **_addOption_** function?
Regards,
Aymeric

Paul-Reed:
Aah right, in that case no worries. Thanks.

On 21 Jun 2017 10:33 am, "Aymeric THIBAUT" <notifications@github.com> wrote:

Hello Paul,
In fact the default colors coded are just defined to create the object in
case initially there are void values (just a security). Actually I don't
know how to pass default colors to pulldown lists. Is there somewhere a
documentation about *addOption* function?
Regards,
Aymeric

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/emoncms/dashboard/pull/99#issuecomment-310024616>, or mute
the thread

Paul-Reed:
Thanks Aymeric

chaveiro:
Nice @thib66, thanks for quick response!

PR #101, opened 2017-06-22
Update bar_render.js

thib66:
optimization of the code

### DIFF
--- a/widget/bar/bar_render.js
+++ b/widget/bar/bar_render.js
@@ -137,7 +137,7 @@ function draw_bar(context,
     units_string = units_string || "";
     title = title || "";
     fstyle = fstyle || "2";
-    fweight = fweight || "1";
+    fweight = fweight || "0";
     unitend = unitend || "0";
     colour_label = colour_label || "000";
     static_offset = 1*static_offset || 0;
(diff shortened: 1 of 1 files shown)


thib66:
Before cf92d7d changes :
before
After cf92d7d changes : 
after

Paul-Reed:
Hi Aymeric.
1) Slight issue when updating existing bar widgets is that this PR introduces Font Weight, and the default option appears to be bold. This means that all user's existing bar widget text would change to bold, instead of remaining as 'normal'. Can the default be changed to 'normal'.
2) There appears to be a text overlap when 'Graduations = On', especially when the value is of few digits.

bar


Paul

thib66:
Hello Paul,
could you tell me why arial font is displayed twice in the Fonts menu?
Thanks a lot for your answser.
Regards,
Aymeric

Paul-Reed:
No idea, possible a typo.
I have however noticed that neither Times or Times New Roman are not included in the fonts list which is a little strange as it's a commonly used serif typeface & websafe, so maybe the second arial entry should have been Times or Times New Roman???

Paul

thib66:
image
The screenshot shows the font options for the following widgets : 
- dewpoint
- feed value
- frost point
- heat index
- humidex
- signal
- sun
I think that we have to homogenize the fonts list for all the widgets. What is your opinion? And of course we could add Times or Times New Roman fonts

Paul-Reed:
I think it's Helvetica Neue that's missing from this widget, and a second 'arial' has been added in error.

I agree that the fonts list should be homogenized for all the widgets for consistency, but maybe just align this PR to the other widget font lists, and then look at the overall fonts list as a separate issue??
...and perhaps rather than just my opinion, we could gather views from other users via the forum?

It's unfortunate that the font selection/style/weight isn't contained within a global file, and called from each of the widgets, to avoid duplicating it in every widget and help future consistency.

Paul-Reed:
The spacing at the bottom of the value is very tight. The screenshot below shows Georgia typeface cut off at the bottom.
Do you think that the label and value would look more balanced if it was left aligned, instead of centered?? 

spacing

Paul-Reed:
Aymeric, the alignment looks much better now, but the spacing/padding between and around the label & value still is very tight. The label is almost touching the chart frame and there is a slight overlap between the label and value.
Also, when updating the widget, the font style still defaults to bold instead of normal (see my comment above dated 29th June). It would be great if we could maintain backwards compatibility, and avoid users having change their widget settings wherever possible.

spacing

thib66:
Hello Paul, in order to reproduce the last screenshot you sent, could you tell me which font did you use?
I do not understand why the font style still defaults to bold, the syntax **_fweight = fweight || "0";_** is used to set normal font for "legacy" widget created in the past. Are you sure that you have completly refreshed your browser?
Regards,
Aymeric

Paul-Reed:
OK, starting off with dashboard master, create a bar chart;

  


Then apply the diff of your PR, reboot, and open the dashboard in a chrome incognito window;

  


As you can see 'bold' is selected in the configuration (although the displayed typeface is not bold??)
Even if you now create a new bar chart, in the configuration options, bold is selected by default.

Paul-Reed:
Is it because the combo box is displaying the first option - bold on line 74 ?

`    var fweightoptions = [`
`                    [1, "Bold"],`
`                    [0, "Normal"]`
`                ];`

thib66:
Paul, yes i guess the combo box is displaying the first option if an option value has not been specified in the past.

29a4eff should fix the issues

Paul-Reed:
Yes it does, and the spacing/padding is now much better.
Thanks for your patience!!

Paul

PR #102, opened 2017-06-23
Update led_render.js

thib66:
code optimization

### DIFF
--- a/widget/led/led_render.js
+++ b/widget/led/led_render.js
@@ -27,31 +27,6 @@ function led_widgetlist()
   return widgets;
 }
 
-function led_init()
-{
-  setup_widget_canvas('led');
-}
-
-function led_draw() {
-  $('.led').each(function(index)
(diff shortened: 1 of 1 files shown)


thib66:
Hello,
@TrystanLea, do we need to maintain specific functions for IE8? I tried to display the widget led in Internet Explorer 8 on a virtual machine without succes. According to several web sites, circle.arc is not natively supported by IE8. In the past did you use excanvas.js or a similar library in order that IE8 displays HTML5 canvas?
Moreover, it seems that draw_binary_led and draw_binary_led_ie8 are no longer used. Could you confirm that I can delete those functions?
Regards,
Aymeric

glynhudson:
Thanks @thib66 . I've just spoken to @TrystanLea and can confirm he's happy for IE8 specific functions to be removed. IE8 is no longer supported by Emoncms.

PR #106, opened 2017-07-05
led widget : new style

thib66:
obsolete and unused draw_binary_led suppressed
add an option with 2 choices of style :

    actual style with colour gradients
    style without colour gradients in order to have a flat design (in fact function developed in the past for IE8, but probably this have never worked for this browser ...)


Paul-Reed:
Thanks Aymeric

PR #107, opened 2017-07-06
Update battery_render.js

thib66:
optimization of the code


Paul-Reed:
Thanks again Aymeric!

PR #108, opened 2017-07-08
add options to the widget

thib66:
optimization of the code

### DIFF
--- a/widget/sun/sun_render.js
+++ b/widget/sun/sun_render.js
@@ -37,6 +37,16 @@ function sun_widgetlist(){
     }
   };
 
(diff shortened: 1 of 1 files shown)


thib66:
before b91ec32 : 
sun
the widget is truncated if wrong height or weight is set.
with b91ec32 : 
sun2

thib66:
Hello,
please could you also have a look at #93?
Thanks in advance.
Regards,
Aymeric

Paul-Reed:
Thanks Aymeric. The resizing is now much better.
I've left a comment in #93 

Paul

PR #109, opened 2017-07-08
fix label issue displaying "needle position"

thib66:


Paul-Reed:
Thanks Aymeric.

Paul

PR #110, opened 2017-07-08
Decimals fix

thib66:
add consistency across the widgets, and remove the non-significant digits when decimals are set as automatic


Paul-Reed:
Thanks Aymeric

Paul

PR #112, opened 2017-07-09
add min max position to the dial

thib66:


thib66:
Before 2b97f9c : 
without-min-max
After 2b97f9c : 
with-min-max

thib66:
image

Paul-Reed:
Thanks Aymeric, a nice and useful addition.
I was thinking how we could display the actual min and max values, but it then starts to look very congested. I'm sure you've considered the same.

Paul

thib66:
Hello Paul,
I tried to display min and max values discreetly in the graph, there is perhaps another way to diplay this information.
If you have any idea, i will try to integrate another representation in the widget.
Regards,
Aymeric

Paul-Reed:
Thanks, would it be possible to use a tooltip, so the value is only shown on mouseover?
...just a thought!

Paul

thib66:
Hello Paul
yes the functionality is available with #115.
Enjoy!
Aymeric

PR #113, opened 2017-07-09
add min and max option to the bar widget

thib66:


thib66:
Before c7c1113 : 
image
After c7c1113 : 
image

Paul-Reed:
Hi Aymeric, noticed that the 'min - max bar' colour defaults back to black if the widget configuration is re-opened.
To re-create, change the colour of the min-max bars to for example green, save changes, save dashboard.
Re-open the widget configuration, and the min-max bars have defaulted back to black.

Paul

thib66:
Hello Paul,
Should be fixed.
For your information, I've discovered that the colour option name must begin with "colour", otherwise the colour is ignored in the widget configuration dialog. 
But perhaps, you already knew that.
Regards,
Aymeric

Paul-Reed:
Aymeric, thanks.
No I wasn't aware, but well spotted.

Did you think any further about displaying the min & max values?
Looking at the bar graph now, it's the only widget that has a tooltip displaying the widget title on mouseover, ....but I can't see in the code how that has been implemented....
If the tooltip could display the min/max value when mouseover the respective indicator, that would be a great addition.
By example, see the 'photoshopped' image below.

tooltip

thib66:
Paul,
I think the tile is displayed on mouseover because tiltle option name begins with "title" it is probably hard coded.
Regards,
Aymeric

PR #114, opened 2017-07-10
add module.json to describe module and display version number 

glynhudson:
Discussion: https://community.openenergymonitor.org/t/module-versioning-development/4520/3?u=glyn.hudson

### DIFF
--- a/module.json
+++ b/module.json
@@ -1,0 +1,11 @@
+{
+    "name"         : "Dashboard",
+    "version"      : "1.0.2",
+    "description"  : "Dashboard module for emoncms",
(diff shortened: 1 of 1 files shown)


Paul-Reed:
Should this meta file be called `package.json` instead of `module.json` to widen it's future usage.
If you get to the stage of publishing/installing modules via npm packages, a package file will be required to hold the meta data of the package in this format.
See https://docs.npmjs.com/files/package.json
Regardless of it's title, it can still be parsed for versioning.

Paul

Paul-Reed:
Now added via 841bff50e7ab1f67e648aa58776326d9d0ebf020

PR #115, opened 2017-07-15
add tooltip for min and max values

thib66:


Paul-Reed:
Wow, that was quick!

It doesn't display the min or max values though, only the current value.
Also, the position of the tooltip just cuts across the white border of the dial

tooltip

thib66:
Paul
Yes, the only way to avoid that the tootip cut the white border is to enlarge the dial in its width as the clearRect function is used.
Regards,
Aymeric

thib66:
Paul,
may you test again?
Thanks,
As tooltip is placed at the position 0, 0 of the canvas, if the tooltip cut the dial, please enlarge the widget.
This is the unique way to solve the problem. Please also check that your dial widget is on the top layer if you have several widgets in order to have the ability to display min and max value anyway. In other terms avoid overlapped widgets.
Regards,
Aymeric

Paul-Reed:
It's displaying the min - max values ok now, and the background shade to the value looks much better.
However there are still some issues around the tooltip, as it doesn't seem to clear fully on mouseout - see image below. (is there tooltip styling in bootstrap?)

tooltip

thib66:
Paul, 
a tooltip library exists with the actual version of bootstrap. Which version is used with emoncms?
Regards,
Aymeric

Paul-Reed:
I believe it's v2.3.2
(old I know!)

thib66:
Paul,
I have used another method in order to display tooltip in the widget bar see #116 (without using clearRect function). Please have a look.
Aymeric

Paul-Reed:
I'm assume that you are seeing the same issues as me.
IMO the best option would be to display the value beneath the cursor, and in the same format as the 'Temp' tooltip.

bar

thib66:
yes I have the same issue. it occurs when only the canvas is refreshed and not the whole page.

Paul-Reed:
Well refreshing the whole page...

bar

thib66:
the best option would be to display the value beneath the cursor, but we have to consider potential issues when the cursor is closed to the border of the canvas... What would be the behaviour of the tooltip ? truncated ?

Paul-Reed:
Isn't the position of the tooltip in bootstrap changeable - https://www.w3schools.com/bootstrap/bootstrap_tooltip.asp

Paul

thib66:
It seems that we need to create anchors and HTML tags inside the document. Do you think that using document.write javascript function can solve the problem?

Paul-Reed:
We probably need advice from someone like @chaveiro who has a much better grasp of the system.
I can't help but think that there is a simpler solution to this...

Paul

Paul-Reed:
Would [something like this](https://stackoverflow.com/questions/9956958/changing-the-position-of-bootstrap-popovers-based-on-the-popovers-x-position-in) help, so the tooltip positioning is dynamic?

Paul

thib66:
hello,
my problem is not to have a dynamic position of the tooltip but my problem is the persistance of the tooltip when the canvas is refreshed:
gif
Regards,
Aymeric

Paul-Reed:
I posted that link because I thought that you were looking at using bootstrap, but you were concerned about the tooltip being truncated.

If the tooltip function is available via bootstrap, I think it would be desirable to use it rather than using alternative code.

The 'title' attribute (which we discussed earlier) does seem to function well, and does clear ok on mouseout, and I believe is using the installed bootstrap library.

thib66:
Paul the title is not coming from bootstrap but directly in the declaration of the widget : 
image

Paul-Reed:
I don't know if it helps, but [this is the bootstrap tooltip javascript](https://github.com/emoncms/emoncms/blob/master/Lib/bootstrap/js/bootstrap.js#L1051) (as currently available from within emoncms).
I've just been trying to implement it into the chart [with help from this](https://www.w3schools.com/bootstrap/bootstrap_tooltip.asp), but my lack of coding skills is letting me down (again!)

Paul

thib66:
the tooltips for dial and bar widgets are similar to those ones used with rawdata graphs : 
image

thib66:
Paul,
unable to use correctly bootstrap tooltip javascript as I am unable to pass variable in the declaration for instance : 
		**_var ttname="\""+"can-"+canvasid+"+2tt"+i+"\"";
		var titletip = h.tip;
        $("[tooltipname="+ttname+"]").tooltip('hide')
          .attr('title', titletip)
          .tooltip('show');_**
seems not working.
I am able to diplay tootips hard coded without variables in an emoncms dashboard.
I think the method actually used for dial and bar widgets without using boostrap tootlip library is acceptable as the graphs are no longer cutted and the behaviour of the tooltips is now correct once the canvas refreshed.
Moreover, the look and feel of the actual tooltips is similar to rawdata graphs one.
Let me know if the issue is closed.
Regards,
Aymeric

thib66:
Actual  render : 
image

Paul-Reed:
Hi Aymeric, thanks.
I'm not home tonight, but just had a quick look, and noticed that you have made changes to the same file - `dial_render.js` in two separate PR's - #115 and #116 

It would be great if we had either just one PR, or two separate PR's - one for the dial, and another for the bar, but changing the same file from 2 different PR's will inevitably cause errors when trying to merge the second PR as there are conflicting lines of code in both.

Paul

thib66:
Hello, yes i done a mistake for #116. How Can I delete the dial-render.js whithin the PR?
Regards,
Aymeric

thib66:
Hello Paul I think I have solve the trouble. I have modified #116 by applying the original content coming from master in the dial-render.js. Now github indicates that only one file has been changed.
Please let me know if this is a good method to reverse a commit.
Regards,
Aymeric

Paul-Reed:
Hi Aymeric
The tooltips looks great!
I've just noticed though that the min/max indicators do not move unless the page is refreshed - same result in bar widget.
To test, set up a dial widget using min & max feeds, and using the api change the values of min feed (in the below screenshot the actual feed values are displayed to the left).

Paul

test

Paul-Reed:
Nice work Aymeric!

PR #118, opened 2017-07-22
add option and better render

thib66:


thib66:
Hello,
before 0116c25 : 
bar_old
When value is higher that max value, the cap is hidden and value overshoots the top of the battery.
bar_old2
The cap is not correctly visible especially if the background of the dashboard is white

thib66:
After 0116c25 : 
bar_new1
The cap is correctly displayed and as well the lines delimiting the battery
bar_new2
When value is higher that max value, the value no longer overshoots the top of the battery.

Paul-Reed:
Thanks Aymeric, looks much better, especially with an outline to the battery cap (so it can now be seen on a white background).

Paul

PR #123, opened 2017-07-24
Update french localization files

thib66:


Paul-Reed:
Hi Aymeric, thanks again!
Because of the way my system is set up, it's difficult to test different localization's to make sure that they've merged OK and working correctly.
I've read through the changes and all appears OK, but would be grateful if you could confirm all is good in your locale.

Paul

thib66:
image
image
image
image

Hello Paul,
the french localization seems to be OK as shown in the screenshots above.
Regards,
Aymeric

Paul-Reed:
Great! thank you.

Paul

PR #124, opened 2017-08-05
Create thermometer_render.js

thib66:
image

### DIFF
--- a/widget/thermometer/thermometer_render.js
+++ b/widget/thermometer/thermometer_render.js
@@ -89,8 +89,8 @@ function thermometer_widgetlist()
                     [1, _Tr("Yes")]
                     ];
 
-    addOption(widgets["thermometer"], "title_thermometer",      "value",            _Tr("Title"),           _Tr("Title of thermometer"),                                                                []);
-    addOption(widgets["thermometer"], "colour_label",   "colour_picker",    _Tr("Label Colour"),    _Tr("Colour of title and values"),                                                  []);
+    addOption(widgets["thermometer"], "titleThermometer",      "value",            _Tr("Title"),           _Tr("Title of thermometer"),                                                                []);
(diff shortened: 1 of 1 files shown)


Paul-Reed:
The codacy score is let down by most of the variables not utilizing Camelcase.
For example, using `displayValue` instead of `display_value`

Paul

thib66:
Hello Paul,
the code should be better.
is there a way to indicate to codacy to ignore undefined **__Tr_** function? (used for localization).
div1 and div2 are tag objects i do not know if we can declare those objects as variables.
concerning the functions I prefer to keep underscores as this particular syntax is used by the emoncms structure.
Regards

Paul-Reed:
That's much better, thanks.
I'm not sure that the tag objects need to be declared, but the parentheses doesn't look quite right.
Are tag objects normally <enclosed> ?? 

Paul

thib66:
helloPaul,
tag objects are normally enclosed <>, if you use jquery this is the case but not when using document.createElement.
Regards,
Aymeric

Paul-Reed:
Aymeric, another useful widget, many thanks.

Paul

Paul-Reed:
Aymeric, any ideas about this issue created in the forum - https://community.openenergymonitor.org/t/new-thermometer-widget-added-to-dashboard/4752

thib66:
Hello Paul,
bug not reproduced with FireFox 54.0.1 and Edge. But reproduced with IE11 as described.
Regards,
Aymeric

thib66:
Paul, 
I have improved the js scripts yesterday for bar, dial and thermometer widgets with declaration of the variables for the tooltips. I do not have errors with those new scripts in IE11.
But #126 should be validated before destroying and creating again my fork in github in order to correctly commit the new fixes. Could you have a look a this PR?
Regards,
Aymeric

PR #126, opened 2017-08-05
Localization improvements

thib66:
image
image


glynhudson:
Thanks a lot for this, could you take a look at the Codacy recommendations?

thib66:
Hello,
the new commits should improve the code.
Concerning _Expected method shorthand_, the syntax seems to be interpreted only with IE11 and later since this is a ECMAScript 6 syntax. Should we apply the codacy recommendations?
Regards,
Aymeric

Paul-Reed:
Looks much better, thanks Aymeric.
To ensure a wider compatibility, let's leave the 'Expected method shorthand' for now.

Paul

PR #128, opened 2017-08-09
Tooltips improvements

thib66:
Better management of tooltips divs

### DIFF
--- a/widget/bar/bar_render.js
+++ b/widget/bar/bar_render.js
@@ -310,6 +310,11 @@ function draw_bar(context,
       var ch=canvas.height;
       var offsetX,offsetY;
       var mouseX,mouseY;
+      var h;
+      var dx,dy;
+
+      var tt1 = document.getElementById(canvas.id + "-tooltip-1");
+      var tt2 = document.getElementById(canvas.id + "-tooltip-2");
 
       function reOffset(){
        var BB=canvas.getBoundingClientRect();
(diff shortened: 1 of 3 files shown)


thib66:
Paul,
could you test the changes in IE11 or ask to Andi to do the tests with its browser?
Regards,
Aymeric

Paul-Reed:
I'm away from home (and a computer!) until tomorrow evening,  unless Glyn reviewes & merges in the meantime.
Thanks for a quick response.

Paul

back2live:
Hello, 
now i don´t get an error, but the widgets want load, are not visible.

Andi

thib66:
Hello Andi,
My widgets are visible, here is a screenshot about a thermometer widget : 
image
Please could you send here a screenshot about your issue?
Is all your widgets not visible or only the widgets containing min and max values?
Also have you cleaned the cache of your browser and click F5 after applying the 3 new js files?
Thanks a lot for your answer.
Regards,
Aymeric

back2live:
Sorry my mistake, i just updated my master. With these 3 new files everything is working fine. Thank you

thib66:
Hello, 
Good news ! Thanks for your feedback.
I used to test my scripts only with Firefox. I've just noticed that there is some warnings in the E11 console.
I will try to fix those issues in a separate PR.
Thanks again for your help.
Regards,
Aymeric

thib66:
Hello,
#129 fixes most of warnings previously displayed in IE11 console.
Aymeric

Paul-Reed:
Aymeric, someone has just raised a question in the forum, "does the thermometer handles negative temperatures", and checking, it doesn't go any lower than zero!!
(I can't believe I missed that when testing)
Should the minimum temperature be set by the users in the same way as the maximum temperature, and of course functionality for the minimum bar to move into the negative.

thib66:
Hello Paul,
yes the thermometer handles negative temperatures, you need to use the offset option and change the max value option.
For instance if you decide that the temperature min is -20°C and the temperature max is +30°C.
Then for the **_offset_** set **_-20_** and for the **_max temperature_** set **_50_** as shown below : 
image
As a result you get this : 
image
image
Let me know if you need further information.
Regards,
Aymeric

thib66:
image

Paul-Reed:
Yes it works by doing that, but I don't think users will find it very intuitive, especially as the Max value will be set at 50 whilst the max value displayed will just be 30.
I realize that negative numbers are handled differently in different widgets, but the jguage widget handles this very well, by setting both min & max in the settings, and doing the offset in code.

jguage  

jguage2

Paul-Reed:
Thanks for the fix Aymeric

PR #129, opened 2017-08-10
Fix syntax widget view

thib66:
fix syntax errors detected with IE11 (but not detected with codacy)


Paul-Reed:
Nice work Americ, thank you.

PR #131, opened 2017-08-19
Tooltip improvement

thib66:
Add z-index property in order to be sure that tooltip is visible when the tooltip is outside the area of the associated widget


Paul-Reed:
Thanks Aymeric.

Paul

PR #133, opened 2017-08-29
add min option for the widget

thib66:
As requested in PR #128 , a minimum option has been added


thib66:
if the improvement gives satisfaction, please add a comment for me in : 
https://community.openenergymonitor.org/t/new-thermometer-widget-added-to-dashboard/4752/5
in order to indicate that min option is now available.
Thanks,
Aymeric

Paul-Reed:
Thanks Aymeric, I'm sure that users will appreciate the change.
I'll update the forum post.

Paul

PR #137, opened 2017-09-06
Better management of update

thib66:
code added in order to correctly update min and max values coming from virtual feeds


Paul-Reed:
Thanks Aymeric.
I've just posted https://github.com/emoncms/dashboard/issues/139 which is a minor glitch with the thermometer widget.

Paul

PR #141, opened 2017-09-10
thermometer base improved

thib66:


Paul-Reed:
I see that you are revoking the previous change to the min/max redraw criteria in this PR, is that intentional?

Paul

thib66:
Hello Paul,
it was not intentional. Should be fixed with 7e2473e.
Aymeric

Paul-Reed:
Thanks Aymeric.
Are you picking up any browser console errors from the dial widget?

Paul

thib66:
Hello, 
yes there is an error displayed by the IE11 console indicating that there is an issue with the </body> tag.
Looking at the source I have a <body> tag in the top and a </body> tag at the bottom. So I do not see where is the issue. Maybe a confusion with the parser of IE as body is also present in : $('body').css("background-color","#ffffff");
Aymeric

Paul-Reed:
I'm not home right now, but the error which I got pointed to L484 (the last
commit) of dial_render.php.
Could you please check the structure of the 'if' statement.

Thanks
Paul

PR #142, opened 2018-01-22
Added timeout in FeedValue, Dial and Bar and new Widget called FeedTime

EA2EGA:
Hi all, this is my first pull request to th EmonCMS project.

For I time, I had the problem of not noticing when a device stopped sending data, because the widgets continued to display the feed value, even if it wasn't updated.

The 3 widgets I use commonly are FeedValue, Dial and Bar, so I edited only these.

I have added a new parameter in the config options of each widget, called timeout. After the seconds specified in timeout without feed update, the widget shows "TO Error" in the numeric display and the Dial or the Bar goes to 0.

If 0 is specified in the timeout value, it never timeouts. If it is left empty, it timeouts to 60 seconds. I added this last feature for not changing all the dashboards I have.

I also added a new widget called FeedTime, which shows how old is a feed in seconds.

Thanks :)

### DIFF
--- a/widget/bar/bar_render.js
+++ b/widget/bar/bar_render.js
@@ -108,7 +108,7 @@ function bar_widgetlist()
     addOption(widgets["bar"], "minvaluefeed",   "feedid",           _Tr("Min. feed"),       _Tr("The feed for the minimum value"),                                              []);
     addOption(widgets["bar"], "maxvaluefeed",   "feedid",           _Tr("Max. feed"),       _Tr("The feed for the maximum value"),                                              []);
     addOption(widgets["bar"], "colour_minmax",  "colour_picker",    _Tr("Colour"),          _Tr("Colour for min. and max. bars"),                                               []);
-    addOption(widgets["bar"], "timeout",        "value",            _Tr("Timeout"),         _Tr("Timeout without feed update in seconds (0 is never, empty defaults to 60)"),   []);
+    addOption(widgets["bar"], "timeout",        "value",            _Tr("Timeout"),         _Tr("Timeout without feed update in seconds (empty is never)"),   []);
 
 
 
--- a/widget/bar/bar_render.js
+++ b/widget/bar/bar_render.js
@@ -140,7 +140,7 @@ function draw_bar(context,
                   minvaluefeed,
                   maxvaluefeed,
                   colour_minmax,
-                  error_code
+                  errorCode
                   )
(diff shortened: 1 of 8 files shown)


Paul-Reed:
Hi, and thanks for submitting.

1) Would it be possible to add a `Title` field to the FeedTime widget? That would help users identify which feed the widget is actually monitoring.
2) To help improve code quality (Codacy score), could you please use CamelCase wherever possible. For example; draw_gauge, feedtime_widgetlist, error_timeout, preferable should be drawGauge, feedtimeWidgetlist, errorTimeout. See https://www.codacy.com/app/emoncms/dashboard/pullRequest?prid=1257061
3) Thinking about 'Backwards compatibility'... would it be better if 'empty defaults to never' instead of '60'? Many of my feeds only update every 2 - 5 minutes by design, so if I updated, by default I would get the `TO Error` message on my existing dashboard widgets. If I wanted '60', I of course would have the option to select it. What do you think?

Paul

EA2EGA:
Hi Paul,

I'll try to fix those things :)

Thanks,



2018-01-27 22:48 GMT+01:00 Paul <notifications@github.com>:

Paul-Reed:
Thanks. Please note that I edited my post to not include the addition of a name Field. I added the field myself, but it restricted how the widget could be displayed in a dashboard, and just didn't look good. So probably better without it.

glynhudson:
>I have added a new parameter in the config options of each widget, called timeout. After the seconds specified in timeout without feed update, the widget shows "TO Error" in the numeric display and the Dial or the Bar goes to 0.

This is a really great idea. Thanks for the PR. Would it be possible to add a screenshot to illustrate operation?

Paul-Reed:
The dial has it's timout set to 10 seconds, and the new FeedTime widget is below, showing the time elapsed since the feed was last updated.

Paul

timeout

TrystanLea:
Looks great! merge when your happy

EA2EGA:
I have made some improvments to code quality, but it seem that I don't pass the Codacy test yet.
I won't be able to review the code for two or more weeks so I left it as is.
Thanks,

Paul-Reed:
Thanks Xabier, a nice addition.
Realise that you are busy for a couple of weeks, but it would be great if this could be extended to the other widgets too, so that there is a common approach to handling inactive feeds.

Regards

Paul

PR #146, opened 2018-03-09
fix ie11 error

thib66:
image

Comma removed. Without this fix, widgets are not displayed in IE11 (works with firefox and Edge anyway)


glynhudson:
thanks 👍

PR #147, opened 2018-03-09
Add double click action on widget

thib66:
By directly double clicking on a widget, the widget options popup appears.

### DIFF
--- a/Views/js/designer.js
+++ b/Views/js/designer.js
@@ -619,6 +619,13 @@ var designer = {
                 }
             }
         });
+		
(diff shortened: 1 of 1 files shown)


glynhudson:
Nice! Thanks works well 👍 

Just a thought I wonder how easy it would be to increase the size of the little box in the middle of the widget the user needs to click and hold to drag the widget? It's quite small and hard to click. The same goes for resizing.

thib66:
Hello @glynhudson,
please see #148.
Regards,
Aymeric

PR #148, opened 2018-03-11
dynamic size of box for resizing and moving of widget

thib66:
image
image
image
if the width and the height of the widget are greater than 125, the size of the boxes is set to 25
if the width and the height of the widget are greater than 75, the size of the boxes is set to 16
otherwise the size of the boxes is set to 8 (legacy size)


glynhudson:
Thanks great, much easier to move around.

Thanks 👍

PR #150, opened 2018-03-24
 Fix for bar min/max feed value

thib66:
hide feature but not bar when feed ids are invalid


Paul-Reed:
Could you further describe what this PR achieves please.

Paul

thib66:
Hello, suppose you have defined min and max options with existing feed.
Then, if inadvertently one of the feeds are deleted (or all), you get the following errors in the FireFox console : 
image
This PR avoids this kind of error and disable the min max feature until you modify the options to choose an existing and valid feed(s)
Let me know if you need further information,
Aymeric

PR #152, opened 2018-03-24
 Fix for dial min/max feed value

thib66:
hide feature but not dial when feed ids are invalid


Paul-Reed:
Thanks Aymeric, all three PR's merged into the 'minmaxfix' branch for testing.

Paul

PR #153, opened 2018-03-25
min-max fix

Paul-Reed:
Hides min-max feature in the event of invalid feed (thib66)


Paul-Reed:
The codacy report has been considered, but to use camelcase would create inconsistencies with the existing var structure.

Paul

PR #158, opened 2018-05-21
Localization vis options

thib66:
Before PR : 
image
After PR : 
image

Should be applied in conjunction with #346 of emoncms/emoncms repository


thib66:
Should be applied in conjunction with #846 of emoncms/emoncms repository.

glynhudson:
Thanks! :+1:

PR #162, opened 2018-06-17
enlarge space for tags - color picker improvement

thib66:
Before : 
image
text truncated
image
issue with color picker on FireFox 60.0.2
After : 
image
text no longer truncated : size increased for tags and font size decreased
image
color picker now correctly displayed on Firefox


Paul-Reed:
I hadn't noticed the problem with Firefox until you pointed it out!
Thanks Aymeric.

PR #180, opened 2019-01-04
Js redundancy

thib66:
Actually,
the javacript code used to define the offset is duplicated in the HTML source of a Dashboard : 
image
This PR solves the issue.


Paul-Reed:
*** CAUTION - DO NOT MERGE ***

@thib66 - looking at the proposed changes, this PR appears to delete the widgets from the Dashboard??

thib66:
Hello Paul,
the widgets are still avalaible since the scripts <widget>_render.js are still present in the widget folder.
As a result : 
image
Let me know if you need further information,
Regards

Paul-Reed:
Ah, I see what you've done!
Thanks Aymeric.

PR #182, opened 2019-01-17
Feature/testing generic units

emrysr:
image

this pull request tests the common units feature of  emoncms `emoncms/Lib/units.php` 
re: https://github.com/emoncms/emoncms/issues/1001

this is just one widget (bar widget) with the code in place but if it works as expected I can roll it out to the other widgets that use a `unit` value


emrysr:
have added and tested the new units dropdown with all these widgets:
image

image



the default units can be found in `emoncms/Lib/units.php` of https://github.com/emoncms/emoncms

TrystanLea:
fantastic works well!

PR #188, opened 2019-02-10
signal render timeout improvement

thib66:
When timeout is reached, the previous signal level is hidden and an error message can be displayed


thib66:
here are some screen captures concerning the behaviour : 
when timeout is not reached : 
image
when timeout is reached : 
image
New widget options : 
image

TrystanLea:
sounds good thanks @thib66

PR #189, opened 2019-02-10
Sun widget timeout improvement

thib66:


thib66:
When timeout is reached, the value is replaced by an error message if timeout specified.
here are some screen captures concerning the behaviour :
when timeout is not reached :
image
when timeout is reached : 
image
New widget options :
image

PR #194, opened 2019-02-15
frostpoint - aligment

thib66:
Change canvas to text like in the feedvalue widget in order to avoid blurred values when zooming from the web browser.
Add align option - Change order of options to be consistent with other based text widgets


TrystanLea:
thanks @thib66

PR #216, opened 2019-06-14
Feature/issue 213 dashboard list

emrysr:
re made the dashboards list to use vuejs to update the api as the table items are changed.

delwedd

It is also responsive and now fits on a very small device:
delwedd

also requres emoncms update: https://github.com/emoncms/emoncms/pull/1310


glynhudson:
Thanks, @emrysr :+1: 

Just tested, works great. Thanks a lot!

PR #234, opened 2019-08-30
checked for empty value in dashboard alias

### LINKED ISSUE #233: Unable to view dashboard list
Since updating to emoncms v10 I have not been able to access the dashboard list, latest master branch versions installed and updated, db updated, browser cache cleared.  

image

line 1072 col 53 is v.alias.length

image

emrysr:
fix #233 

thanks @pb66 for the detailed explanation in #233

### DIFF
--- a/Views/dashboard_list.php
+++ b/Views/dashboard_list.php
@@ -623,12 +623,16 @@
                     function(data){
                         // handle success - populate gridData[] array
                         // add urls for edit and view
-                        data.forEach(function(v,i){
-                            let id = data[i].id;
-                            data[i].edit = path + 'dashboard/edit?id=' + id;
-                            data[i].view = path + 'dashboard/view';
-                            data[i].view += v.alias.length > 0 ? '/' + v.alias: '?id=' + id;
-                        });
+                        for(n in data) {
+                            let item = data[n];
+                            if (item) {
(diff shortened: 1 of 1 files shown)


pb66:
Thanks emrysr, am I right in thinking this just allows the grid to populate **despite** null aliases?

So the clones will still be created with a null alias whilst original new dashboards are not?

emrysr:
@pb66 - yes the api has not changed. Only the display of the list changed to set the alias to empty when looping through the results.

Do you think the value be set to and `""` (empty string) instead of `null` ?
If so the api endpoint `/dashboard/clone.json` requires a change in the `INSERT` sql to save `""` (empty strings) into the `alias` field.

function `dashclone()` in `dashboard_model.php`
https://github.com/emoncms/dashboard/blob/master/dashboard_model.php#L54

this *should do it on the above line (untested):
```
$this->mysqli->query("INSERT INTO dashboard (`userid`,`content`,`name`,`description`,`height`,`alias`) VALUES ('$userid','{$row['content']}','$name','{$row['description']}','{$row['height']}')", "");
```

I think a new issue should be opened to do the work/tests

emrysr:
this pull request is ready to merge. the above comment can be used to create a new issue if required.

PR #237, opened 2019-09-24
Create kwhdaylastweek_render.js

danbates2:


danbates2:
Working, but had to disable fast_update.

PR #240, opened 2019-10-19
Check for access before including navbar menu

brandock:
If there are no available dashboards and the user doesn’t have a write session then leave the dashboard menu out of the navbar. (It does not serve any function in this situation.)

### DIFF
--- a/dashboard_menu.php
+++ b/dashboard_menu.php
@@ -12,7 +12,12 @@
         }
     }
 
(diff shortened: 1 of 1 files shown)


brandock:
Based on our conversation:
https://community.openenergymonitor.org/t/suggestion-remove-dashboard-sidebar-menu-when-empty/12205/4

PR #243, opened 2019-12-21
kwhperiod_render.js

danbates2:


danbates2:
Not sure why some of the files have changed here.
I made an Auto-Format of the kwhperiod widget, so it looks like a larger diff than it is.

danbates2:
I've rebased my fork. Should be good to go.

widget/kwhperiod/kwhperiod_render.js
```diff
@@ -5,394 +5,849 @@
...
+    } else {
+      // thisperiodEndValue
+      /*
+      var result = feed.get_value(feedid, now.getTime())
+      console.log(result, 'recent period end / now')
+      var thisperiodEndValue = result[1]
+      */
+      // thisPeriodStartValue
+      var result = feed.get_value(feedid, thisPeriodStartTime.getTime())
+      console.log(result, 'recent period start value')
```
emrysr: keep output clean in production. should only output to console during testing

widget/kwhperiod/kwhperiod_render.js
```diff
@@ -5,394 +5,849 @@
...
+      var thisPeriodStartValue = result[1]
+
+      // ... and calculate the result of our time window.
+      console.log(val, 'value now')
```
emrysr: keep output clean in production. should only output to console during testing

emrysr:
do you use the same wiget to compare the previous and current periods??

> eg. showing today's accumulative total as well as yesterday's

I only see one dashboard widget `kwhperiod`...
delwedd

danbates2:
I've commented out the console logs.
Re: question if this is the only widget - YES! It's capable of the Recent Period and Past Period according to the setting 'Periods Ago'.
Did you spot the documentation I made? Ideally I would like to like the documentation in the widget settings dialog box. Any idea how we might do this?

danbates2:
I've found a bug in this. If the data is not available the result falls back on the most recent value of the feed, which is misleading. A Null should be displayed instead I believe.

danbates2:
I believe the bug is now fixed, however it's not a very clever fix. It'll do.

TrystanLea:
Thanks @danbates2!!

PR #245, opened 2019-12-23
added new property to dashboard sql table "fullscreen"

emrysr:
added new checkbox to allow user to set property
added js to alter view template if property set

added new database field. **Need to update database** schema to enable this feature correctly
` EmonCMS->Admin->Update...->Update Database Only`
delwedd

[screenshot of new dashboard option]
delwedd

[screenshot of dashboard edit]
delwedd

[screenshot of same dashboard being viewed without menus showing]
delwedd


chaveiro:
May i recommend that the default behaviour is to have the menu hidden?

PR #260, opened 2020-03-30
Add blocks feature

thib66:
Add the ability to define the number of blocks within the battery : 
image
image
For backward compatibility : default value is 5


TrystanLea:
Thanks a lot @thib66

PR #266, opened 2021-01-03
Fix for 'Undefined variable' on save.

mabi:
Avoid 'Undefined variable' when saving if php-mbstring is not installed.
In turn this notice is sent to the browser causing a JSON parse error.


TrystanLea:
Thanks @mabi

Issue #267, opened 2021-03-04
Recover the published status from v9.8


chaveiro:
Originally you had a setting for each dashboard to consider it a draft or published. 
Only published dashboards were displayed on the menu, very useful to keep tests out of view until ready.
This was lost with recent changes.

Please recover it, only display on the menu (2) the dashboards that are in the publish state (1) and allow user to config status on the dashboard list page (used to be there).

Old UI, keep (1) and only display the (2) links on the new sidebar dashboard menu.
![image](https://user-images.githubusercontent.com/4527135/109943060-b2362e80-7ccc-11eb-82fd-33155f10772e.png)

PR #268, opened 2021-03-11
Recover publish state from v9.8.

chaveiro:
Only display published dashboards on the menu to keep tidiness.

See https://github.com/emoncms/dashboard/issues/267


TrystanLea:
Thanks for this @chaveiro my menu_v3 branch is soon replacing this menu file, if you can rebase and test against the menu_v3 branch that would be great.

TrystanLea:
Here's the menu_v3 branch https://github.com/emoncms/dashboard/tree/menu_v3

